PR #1, opened 2017-07-12
[WIP] Implement methods for bridge between GUI and Opal

theiostream:


Source/opal/OpalBridge.m
```diff
@@ -0,0 +1,126 @@
...
+
+#import <AppKit/NSColor.h>
+#import <AppKit/NSImage.h>
+#import <AppKit/NSColorSpace.h>
+#import <AppKit/NSGraphics.h>
+#import <CoreGraphics/CGColor.h>
+#import <CoreGraphics/CGImage.h>
+#import <objc/runtime.h>
+
+@interface GSNamedColor : NSColor
```
fredkiefer: All these interfaces need to be removed.

Source/opal/OpalBridge.m
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+  if (cgColorSpaceName == NULL)
+    return NULL;
+
+  CGColorSpaceRef colorSpace = CGColorSpaceCreateWithName(cgColorSpaceName);
+  CGFloat values[10];
+  [self getComponents: values];
+
+  CGColorRef color = CGColorCreate(colorSpace, values);
```
fredkiefer: As I wrote in a mail, this is wrong for the NSCalibratedBlackColorSpace] and NSDeviceBlackColorSpace] colour spaces. I think it would be better to ignore them in the code above.

Source/opal/OpalContext.m
```diff
@@ -168,7 +168,9 @@ - (void) GSSetDevice: (void *)device
 {
   OpalSurface *surface;
 
-  surface = [[OpalSurface alloc] initWithDevice: device];
+  CGContextRef suppliedContext = (CGContextRef)[self _graphicsPort];
```
fredkiefer: This method is gone now and this is a good thing :-)

Source/opal/OpalContext.m
```diff
@@ -222,4 +224,16 @@ - (void *) graphicsPort
   return [surface CGContext];
 }
 
+- (CGContextRef) CGContext
```
fredkiefer: We could have these two methods in gui.

Source/opal/OpalSurface.m
```diff
@@ -104,7 +104,7 @@ - (void) createCGContexts
       _gsWindowDevice->gdriverProtocol |= GDriverHandlesExpose | GDriverHandlesBacking;
       _gsWindowDevice->gdriver = self;
 
-      _backingCGContext = createCGBitmapContext(
+      _backingCGContext = ctx ?: createCGBitmapContext(
```
fredkiefer: Shouldn't this read:

 _backingCGContext = ctx ? ctx : createCGBitmapContext(

I also think that we should rather set the _x11CGContext as this is the real context drawings end up in.

theiostream:
@ivucica could you review this and possibly merge?

Source/opal/OpalBridge.m
fredkiefer: @theiostream Daniel you never replied to this comment.

ivucica:
@theiostream Please address comments from Fred first.

(I wish GitHub comment threads had an 'addressed' status, and that they fully preserved old commits until the PR was merged. I wish I could compare files as the PR progressed. But eh.)

Source/opal/OpalBridge.m
theiostream: I believe I added:

```
// FIXME: we should handle black color spaces here, which we currently
// ignore in the implementation.
```

ivucica: Yeh, please just point these things to us :)

GitHub makes it hard to track resolved and unresolved comments. :)

ivucica:
I'll wait for Fred's approval as I don't have the understanding of the problem area that Fred has.

Source/opal/OpalBridge.m
fredkiefer: That new comment is wrong. You don't ignore the black colour spaces you just handle them wrong. I will clean that up as soon as this change gets merged.

PR #5, opened 2019-03-26
Font system revamp

Deek:
This is a semi-comprehensive revamp of the font system in use with the Cairo backend.

I have implemented support for Fontconfig 2.11's unique PostScript font descriptors, which prevents synthesized font names from causing fonts to "shadow" each other. For people with very old versions of fontconfig, I have added many new rules for synthesizing new font descriptors, which should limit that shadowing effect drastically, as well as made use of each font's human-readable name to improve the user interface (and make shadowing even less likely).

I have also introduced sorting for the font list, which causes the faces within a font to become sorted by weight and traits, so that the user-visible list now has a discernible pattern to it that it lacked before.

I have used Cairo's and Fontconfig's support for the FreeType library to access encoding information and PostScript glyph tables to enable -glyphIsEncoded: and -glyphWithName: to finally work, which enables things like ligatures and (eventually) complex glyph substitutions.

I have also included some stubbed-out code which enables the fetching of TrueType SFNT tables, which are the things that actually provide glyph substitution as well as all of the basic typographic metric data we can use to find any font's true typographic baseline, xHeight, and glyph extents.

And finally, for my own purposes, I turned on subpixel rendering and hinting in a user-controllable way.

Fred asked me to put this together some time ago, but I have been busy and hadn't gotten around to it. Sorry. :)

### DIFF
--- a/Source/art/FTFontEnumerator.m
+++ b/Source/art/FTFontEnumerator.m
@@ -245,8 +245,7 @@ static void add_face(NSString *family, int family_weight,
       return;
     }
 
-  fi->displayName = [[family stringByAppendingString: @" "]
-			     stringByAppendingString: faceName];
+  fi->displayName = [[NSString stringWithFormat: @"%@ %@", family, faceName] retain];
 
 
   weight = family_weight;
--- a/Source/cairo/CairoFontInfo.m
+++ b/Source/cairo/CairoFontInfo.m
@@ -35,6 +35,13 @@
 #include <math.h>
 #include <cairo-ft.h>
 
+#include FT_FREETYPE_H
+#include FT_SFNT_NAMES_H
+#include FT_TRUETYPE_TABLES_H
+#include FT_TYPE1_TABLES_H
+#include <fontconfig/fontconfig.h>
+#include <fontconfig/fcfreetype.h>
+
 @implementation CairoFontInfo 
 
 - (BOOL) setupAttributes
@@ -159,15 +166,38 @@
 
 - (BOOL) glyphIsEncoded: (NSGlyph)glyph
 {
-  /* FIXME: There is no proper way to determine with the toy font API,
-     whether a glyph is supported or not. We will just ignore ligatures 
-     and report all other glyph as existing.
-  return !NSEqualSizes([self advancementForGlyph: glyph], NSZeroSize);
-  */
-  if ((glyph >= 0xFB00) && (glyph <= 0xFB05))
-    return NO;
-  else
-    return YES;
+  BOOL val = NO;
+
+  if (_scaled)
+    {
+      FT_Face face = cairo_ft_scaled_font_lock_face(_scaled);
+      if (FT_Get_Char_Index (face, glyph) != 0)
+	{
+	  val = YES;
+	}
+
+      cairo_ft_scaled_font_unlock_face(_scaled);
+    }
+  return val;
+}
+
+- (NSGlyph) glyphWithName: (NSString *)glyphName
+{
+  NSGlyph val = NSNullGlyph;
+  FT_String *name = (FT_String *)[glyphName UTF8String];
+
+  if (_scaled)
+    {
+      FT_Face face = cairo_ft_scaled_font_lock_face(_scaled);
+      if (FT_Has_PS_Glyph_Names(face))
+	{
+	  val = (NSGlyph) FT_Get_Name_Index(face, name);
+	}
+
+      cairo_ft_scaled_font_unlock_face(_scaled);
+    }
+
+  return val;
 }
 
 static
--- a/Source/cairo/CairoFontInfo.m
+++ b/Source/cairo/CairoFontInfo.m
@@ -241,6 +241,12 @@ BOOL _cairo_extents_for_NSGlyph(cairo_scaled_font_t *scaled_font, NSGlyph glyph,
   return NSZeroRect;
 }
 
+- (NSString *) displayName
+{
+	return [_faceInfo displayName];
+}
+
+
 - (CGFloat) widthOfString: (NSString *)string
 {
   cairo_text_extents_t ctext;
--- a/Source/cairo/CairoFontInfo.m
+++ b/Source/cairo/CairoFontInfo.m
@@ -51,6 +51,10 @@
   cairo_matrix_t font_matrix;
   cairo_matrix_t ctm;
   cairo_font_options_t *options;
+  cairo_hint_metrics_t metrics = CAIRO_HINT_METRICS_ON;
+  cairo_hint_style_t   style = CAIRO_HINT_STYLE_DEFAULT;
+  NSUserDefaults *ud = [NSUserDefaults standardUserDefaults];
+  int hinting = 0, subpixel = 0;
 
   if (![super setupAttributes])
     {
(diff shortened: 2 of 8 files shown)


Source/fontconfig/FCFontEnumerator.m
```diff
@@ -53,6 +53,87 @@
...
+    {
+      return (float) (fcWeight - bottomValue) * (1.0f / (topValue - bottomValue));
+    }
+}
+
+static NSComparisonResult
+sortFontFacesArray(id fontArr1, id fontArr2, void *context)
+{
+  /*
+    Order of array:
```
fredkiefer: This comment looks incorrect to me when compared to the following code.

Source/fontconfig/FCFontEnumerator.m
```diff
@@ -78,50 +159,157 @@ + (Class) faceInfoClass
...
+     This code will allow us to access TrueType/OpenType font tables from
+     fontconfig. If we want to implement features like the so-called "OS/2"
+     table which contains typographic family info ("give me a list of 'script'
+     fonts with slab serifs") and stuff like the font's correct typographic
+     line height and its xHeight, this is where we get that info.
+
+     For more information, see http://www.microsoft.com/typography/otspec/
+  */
+#if 0
+  if (FcPatternGetFTFace (pat, FC_FT_FACE, 0, &ftface) == FcResultMatch)
```
fredkiefer: For now I would prefer to leave this code out.

Source/fontconfig/FCFontEnumerator.m
```diff
@@ -189,23 +391,27 @@ - (void) enumerateFontsAndFamilies
 
               familyString = [NSString stringWithUTF8String: family];
               familyArray = [fcxft_allFontFamilies objectForKey: familyString];
+
               if (familyArray == nil)
                 {
                   NSDebugLLog(@"NSFont", @"Found font family %@", familyString);
-                  familyArray = [[NSMutableArray alloc] init];
```
fredkiefer: I prefer to have non-autorelease objects especially if they are long lived (but also for the very short lived ones).

Source/fontconfig/FCFontEnumerator.m
```diff
@@ -214,75 +420,86 @@ - (void) enumerateFontsAndFamilies
...
 - (NSString *) defaultFixedPitchFontName
 {
-  if ([allFontNames containsObject: @"Bitstream Vera Sans Mono"])
+  if ([allFontNames containsObject: @"DejaVuSansMono"])
+    {
+      return @"DejaVuSansMono";
+    }
+  if ([allFontNames containsObject: @"BitstreamVeraSansMono-Roman"])
     {
       return @"Bitstream Vera Sans Mono";
```
fredkiefer: Here the names do not match.

Source/fontconfig/FCFontEnumerator.m
```diff
@@ -214,75 +420,86 @@ - (void) enumerateFontsAndFamilies
...
+	  {
+		[[allFontFamilies objectForKey: key] sortUsingFunction:fontSort context:NULL];
+	  }
+  }
 }
 
 - (NSString *) defaultSystemFontName
 {
-  if ([allFontNames containsObject: @"Bitstream Vera Sans"])
+  if ([allFontNames containsObject: @"DejaVuSans"])
```
fredkiefer: Nothing against making DejaVu the default font, but I will have to test whether the names match the new results for "name" on my machine and hopefully many others :-)

Source/cairo/CairoFontInfo.m
```diff
@@ -71,16 +82,48 @@ - (BOOL) setupAttributes
       return NO;
     }
 
-  // We must not leave the hinting settings as their defaults,
```
fredkiefer: How do you make sure that the problem mentioned in the comment won't affect the text drawing?

Source/cairo/CairoFontInfo.m
```diff
@@ -71,16 +82,48 @@ - (BOOL) setupAttributes
...
+	    break;
+	  case 4:
+	    style = CAIRO_HINT_STYLE_FULL;
+	    break;
+	}
+    }
+  cairo_font_options_set_hint_metrics(options, metrics);
+  cairo_font_options_set_hint_style(options, style);
+
+  if ((subpixel = [ud integerForKey: @"back-art-subpixel-text"]))
```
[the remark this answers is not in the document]
Deek: History. The default was added by Alex Malmberg in 2002 or so in the libart backend to do subpixel rendering of text. When he added the feature to back-art we publicized it, and as a result it's been added to user-facing software. So, despite extending support for the feature to Cairo, I felt it was inappropriate to arbitrarily rename the option or create a second one.

Source/art/FTFontEnumerator.m
```diff
@@ -245,8 +245,7 @@ static void add_face(NSString *family, int family_weight,
       return;
     }
 
-  fi->displayName = [[family stringByAppendingString: @" "]
-			     stringByAppendingString: faceName];
+  fi->displayName = [[NSString stringWithFormat: @"%@ %@", family, faceName] retain];
```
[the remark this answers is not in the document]
Deek: True enough, but as it was it's clear enough that this was just a bug that doesn't seem to have bitten anyone in an obvious way, since all -string* methods explicitly return autoreleased instances.

Source/cairo/CairoFontInfo.m
```diff
@@ -159,15 +202,38 @@ - (void) dealloc
...
+  FT_String *name = (FT_String *)[glyphName UTF8String];
+
+  if (_scaled)
+    {
+      FT_Face face = cairo_ft_scaled_font_lock_face(_scaled);
+      if (FT_Has_PS_Glyph_Names(face))
+	{
+	  val = (NSGlyph) FT_Get_Name_Index(face, name);
+	}
+
```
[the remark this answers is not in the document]
Deek: The superclass's implementation doesn't do anything useful, and the fallback is provided by knowing that FT_Get_Name_Index() returns zero when the named glyph is not present (or, really, for any error).

Source/cairo/CairoFontInfo.m
Deek: I don't. In fact, I make note that it does. For some reason, __disabling__ the hinting of metrics causes menus to get cut off by one pixel on my setup -- which is actually the opposite of what should happen. If something is going to screw up, it should happen when you enable hinting of metrics, not when you turn it off.

Source/cairo/CairoFontInfo.m
```diff
@@ -71,16 +82,48 @@ - (BOOL) setupAttributes
...
+	    break;
+	}
+    }
+  cairo_font_options_set_hint_metrics(options, metrics);
+  cairo_font_options_set_hint_style(options, style);
+
+  if ((subpixel = [ud integerForKey: @"back-art-subpixel-text"]))
+    {
+      cairo_font_options_set_antialias(options, CAIRO_ANTIALIAS_SUBPIXEL);
+      cairo_font_options_set_subpixel_order(options, CAIRO_SUBPIXEL_ORDER_RGB);
```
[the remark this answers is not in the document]
Deek: In theory yes, in practice nothing uses BGR so it's an 'if' that results in a second call to cairo_font_options_set_subpixel_order() basically never.

Source/fontconfig/FCFontEnumerator.m
Deek: Which comment? The convertWeight() function does do what the comment says it does, convert a number into a fraction accurately representing the numeric position of "fcWeight" within the range between "bottomValue" and "topValue", while enforcing the rule that fcWeight must indeed be within that range. It's for assigning fractional weight values for fonts that do not match any of the pre-defined Fontconfig weight classes -- which turns out to not be a rare thing, after seeing GNUstep assign "medium" weights to fonts that Fontconfig knew were Ultra Light.

The git comment attached to the commit describes what's happening in detail.

Source/cairo/CairoFontInfo.m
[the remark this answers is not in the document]
Deek: OH, you meant for the first method, -glyphIsEncoded: -- The original method just returned YES for everything but five ligature glyphs, which I find extremely silly, while this version returns NO for everything that isn't a valid character in the font's encoding. I'm not sure what kind of fallback might be useful, since the font itself is the only thing that might know if it has a glyph at index N.

Source/fontconfig/FCFontEnumerator.m
Deek: I'm not sure what you're referring to here -- familyArray _isn't_ autoreleased.

Deek: Oh, I see what you are talking about now. The location at which you made your comment threw me off.

Source/fontconfig/FCFontEnumerator.m
Deek: DejaVu is Vera Sans but with many more glyphs, so anywhere the default font is Vera should really prefer DejaVu _if present_, falling back to Vera and then finally to something that looks worse. :)

Source/fontconfig/FCFontEnumerator.m
Deek: You are correct, that is a bug.

Deek:
By the way, I have recently discovered that under certain conditions, the Fontconfig code is extracting human-readable names from some fonts that are encoded in languages other than English. Most fonts don't exhibit the issue, but some Microsoft ones do. I have looked for a useful way around it, but I haven't found one yet. If you see a font named e.g. "Courier New kurzíva", this is the cause. The problem is *not* on our end, it's in Fontconfig and/or the font itself. Actually taking the step to load and use the SFNT tables ourselves is the probable solution.

Source/fontconfig/FCFontEnumerator.m
fredkiefer: The comment I was referring to is the one about the order of the array. This comment states a certain order and the sorting function is enforcing a different one. One of the two sides is incorrect and I would like to have them consistent.

Source/cairo/CairoFontInfo.m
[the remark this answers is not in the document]
fredkiefer: No, instead we should add a new option, with a meaningful name. What we also should do is to use the old name as a fallback in the art backend, so that there both names get supported. Of course the new name also needs to be documented.

fredkiefer:
How to proceed with this pull request? You did not change anything, not even the bug you admitted. Does this mean I will have to adjust this pull request myself? If so I would prefer the merge it piecewise. If that is fine for you, I will start merging next week.

fredkiefer:
I changed my mind and will no merge the whole change request and do the clean up after that. This seems easier for me. Could you also add a pull request for your changes on gui? Without that the changes to the font display names will never show up anywhere.

PR #6, opened 2019-03-26
Set `Utility` window type for NSFloatingWindowLevel.

trunkmaster:
This change also needs change to AppKit/NSWindow.h - waiting for my previous PR being processed.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2019-03-26 Sergii Stoian <stoyan255@gmail.com>
+	* Source/x11/XGServerWindow.m:
+	(setwindowlevel::) Set `Utility` window type for NSFloatingWindowLevel.
+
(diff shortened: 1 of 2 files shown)


fredkiefer:
This change looks definitely correct. The only question is whether by now enough of our supported window managers implement this level correctly. We will see after merging this.

Could you please provide a ChangeLog entry for this change? After that I will merge it.

trunkmaster:
ChangeLog entry was added.
I see at least several options for this:
1. WindowMaker - it has GNUstep.h contains enum with NS*WindowLevel definition. It's not used through WindowMaker's code.
2. My NextSpace's Workspace - contains code from WindowMaker but it can be easily synchronized with GNUstep code changes.
3. Other window managers should support NETWM atoms. It depends on level of NETWM support there.

PR #7, opened 2019-04-02
Handle `-autolaunch YES` application argument

trunkmaster:
Map application icon window without focus switch for applications executed with argument `-autolaunch YES` in WindowMaker environment.

### DIFF
--- a/Source/x11/XGServerWindow.m
+++ b/Source/x11/XGServerWindow.m
@@ -2821,15 +2821,15 @@ static BOOL didCreatePixmaps;
  	    {
               XEvent ev;
 
-              // Inform WM to ignore focus events
+              // Inform WindowMaker to ignore focus events
               ev.xclient.type = ClientMessage;
-              ev.xclient.message_type = XInternAtom(dpy,"WM_IGNORE_FOCUS_EVENTS", False);
+              ev.xclient.message_type = generic.WM_IGNORE_FOCUS_EVENTS_ATOM;
               ev.xclient.format = 32;
               ev.xclient.data.l[0] = True;
               XSendEvent(dpy, ROOT, True, EnterWindowMask, &ev);
               // Display application icon
               XMapWindow(dpy, ROOT);
-              // Inform WM to process focus events again
+              // Inform WindowMaker to process focus events again
               ev.xclient.data.l[0] = False;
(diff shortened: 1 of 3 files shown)


Source/x11/XGServerWindow.m
```diff
@@ -2815,6 +2815,24 @@ - (void) orderwindow: (int)op : (int)otherWin : (int)winNum
...
 	{
+          NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
+          if (op != NSWindowOut && window->map_state == IsUnmapped &&
+              [[defaults objectForKey: @"autolaunch"] isEqualToString:@"YES"])
+ 	    {
+              XEvent ev;
+
+              // Inform WM to ignore focus events
+              ev.xclient.type = ClientMessage;
+              ev.xclient.message_type = XInternAtom(dpy,"WM_IGNORE_FOCUS_EVENTS", False);
```
fredkiefer: In GNUstep back I did remove all the local calls to XInternAtom and get all the required atoms once at the start of the application, see _setupRootWindow. These atoms are defined in the file XGGeneric.h.
The only place my Google search found this specific atom "WM_IGNORE_FOCUS_EVENTS" was in WindowMaker source code. Is it only supported for this window manager? I also didn't find it in my X Windows reference book. Where does this come from?

trunkmaster: Yes it's WindowMaker only atom (defined here: https://repo.or.cz/wmaker-crm.git/blob/HEAD:/src/startup.c#l399).
Anyway this `if (generic.flags.useWindowMakerIcons == 1)` block (starts at line 2808) will be invoked only for WindowMaker. I suppose `useWindowMakerIcons` flag make sense only for WindowMaker.

ivucica:
Would it make sense, then, to just expand "// Inform WM to ignore focus
events" to "// Inform WindowMaker to ignore focus events", making it even
clearer this is a WindowMaker-only atom?

>

trunkmaster:
@ivucica No problem. But it make sense only for commit diff... In comment above this WindowMaker already mentioned. Anyway, I'll make change if code becomes more readable.

trunkmaster:
@fredkiefer, new atom added.
@ivucica, comments edited.

PR #8, opened 2019-04-03
Send double-click on appicon to WindowMaker

trunkmaster:
Old code sent to WM only single-click. 
Receiving double-click on GNUstep appicons allows WindowMaker to perform its own focus change related actions.

### DIFF
--- a/Source/x11/XGServerEvent.m
+++ b/Source/x11/XGServerEvent.m
@@ -472,10 +472,10 @@ posixFileDescriptor: (NSPosixFileDescriptor*)fileDescriptor
             /*
              * We must hand over control of our icon/miniwindow
              * to Window Maker.
-                 */
+             */
             if ((cWin->win_attrs.window_style
-              & (NSMiniWindowMask | NSIconWindowMask)) != 0
-              && eventType == NSLeftMouseDown && clickCount == 1)
+                 & (NSMiniWindowMask | NSIconWindowMask)) != 0
+                && eventType == NSLeftMouseDown)
               {
                 if (cWin->parent == None)
                   break;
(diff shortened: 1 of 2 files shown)


Source/x11/XGServerEvent.m
```diff
@@ -472,10 +472,10 @@ - (void) processEvent: (XEvent *) event
             /*
              * We must hand over control of our icon/miniwindow
              * to Window Maker.
-                 */
+             */
             if ((cWin->win_attrs.window_style
-              & (NSMiniWindowMask | NSIconWindowMask)) != 0
-              && eventType == NSLeftMouseDown && clickCount == 1)
+                 & (NSMiniWindowMask | NSIconWindowMask)) != 0
```
fredkiefer: Something seems to be wrong with the indentation of these two lines. Could you please check?

Source/x11/XGServerEvent.m
```diff
@@ -484,7 +484,8 @@ - (void) processEvent: (XEvent *) event
                 XSendEvent(dpy, cWin->parent, True,
                            ButtonPressMask, &xEvent);
                 XFlush(dpy);
-                break;
+                if (clickCount != 2)
```
fredkiefer: Why don't you check for "clickCount == 1", which would be closer to the old behaviour? This isn#T important but changes the result of a triple click.

Source/x11/XGServerEvent.m
trunkmaster: The indentation is correct. The line 477 aligned to the inner expression (`c` character of `cWin`). 
Line 478 aligned to the brackets of the if () statement. This is how emacs make an indentation of C code.

Source/x11/XGServerEvent.m
trunkmaster: There was some glitches with old -back and -gui for double-clicks. I've just checked it with current code - everything seems correct.

PR #10, opened 2019-04-10
Support for configurable mouse properties

trunkmaster:
Double-click time, move threshold, scroll wheel multiplier, menu button configuration: enable/disable, left/right.

Things which are need to be done in GUI:
- NSUserDefaults should send notification to NSDistributedNotificationCenter on defaults change. This will allow all running applications adjust mouse properties on the fly.
- NSScrollView (NSScroller) should respect "GSMouseScrollMultiplier" value.

### DIFF
--- a/Source/x11/XGServerEvent.m
+++ b/Source/x11/XGServerEvent.m
@@ -354,8 +354,8 @@ posixFileDescriptor: (NSPosixFileDescriptor*)fileDescriptor
   NSUserDefaults *defs = [NSUserDefaults standardUserDefaults];
 
   clickTime = [defs integerForKey:@"GSDoubleClickTime"];
-  if (clickTime < 250)
-    clickTime = 250;
+  if (clickTime < 200)
+    clickTime = 300;
   
   clickMove = [defs integerForKey:@"GSMouseMoveThreshold"];
   if (clickMove < 3)
--- a/Source/x11/XGServerEvent.m
+++ b/Source/x11/XGServerEvent.m
@@ -42,6 +42,7 @@
 #include <Foundation/NSUserDefaults.h>
 #include <Foundation/NSRunLoop.h>
 #include <Foundation/NSDebug.h>
+#include <Foundation/NSDistributedNotificationCenter.h>
 
 #include "x11/XGServerWindow.h"
 #include "x11/XGInputServer.h"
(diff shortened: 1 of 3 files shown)


Source/x11/XGServerEvent.m
```diff
@@ -328,6 +339,52 @@ - (NSPoint) _XPointToOSPoint: (NSPoint)x for: (void*)window
...
+           name: NSUserDefaultsDidChangeNotification
+         object: nil];
+}
+
+- (void) mouseOptionsChanged: (NSNotification *)aNotif
+{
+  NSUserDefaults *defs = [NSUserDefaults standardUserDefaults];
+
+  clickTime = [defs integerForKey:@"GSDoubleClickTime"];
+  if (clickTime < 250)
```
fredkiefer: Shouldn't we use a lower number as the limit here, e.g. 100, and have the old value 300 as default?

Source/x11/XGServerEvent.m
```diff
@@ -455,6 +528,9 @@ - (void) processEvent: (XEvent *) event
             break;                /* Unknown button */
           }
 
+        if (menuButtonEnabled == NO && eventType == menuMouseButton)
```
fredkiefer: Shouldn't me add this further up where we decide which mouse event to send? we would need to duplicate the code and I am really not sure what is less confusing.

Source/x11/XGServerEvent.m
```diff
@@ -511,14 +587,30 @@ - (void) processEvent: (XEvent *) event
         [self setLastTime: xEvent.xbutton.time];
         if (xEvent.xbutton.button == generic.lMouse)
           {
-            eventType = NSLeftMouseUp;
-            buttonNumber = generic.lMouse;
+            if (swapMouseButtons)
```
fredkiefer: When menuButtonEnabled is set to no, we could get release events without corresponding press events before them. We should add the check from above here as well.

Source/x11/XGServerEvent.m
trunkmaster: From the UX point of view 100 is too low number. The human reaction time is 270-280 milliseconds on average (look here for example: https://www.humanbenchmark.com/tests/reactiontime/statistics). That's why 250 ms. I've tried to set 100 ms and I must say that this is too little time between clicks. Practically we shouldn't allow to set such low number to prevent user annoyance.

Source/x11/XGServerEvent.m
trunkmaster: What does confuse you here? I'd rather not insert return statement into if...else... block of code. Here return statement has good visibility even with code folding editor feature is used. Maybe I got you wrong?

Source/x11/XGServerEvent.m
trunkmaster: You're right. Thanks. Fixed.

Source/x11/XGServerEvent.m
fredkiefer: If 100 is to low, then some other sensible value like 200, but the important bit is to keep the default of 300. So if somebody won't set this value she will get the previous behaviour.

trunkmaster: OK, changed.

PR #13, opened 2019-12-11
Fix focus handling for a number of cases for WindowMaker

trunkmaster:
During NEXTSPACE development I've managed to fix most (if not all) of focus related problems.
Most tricky part was to make reliable focus handling on workspace switch.
Most part of the changes are related to the TakeFocus atom that is specific to WindowMaker, AFAIK.

### DIFF
--- a/Source/x11/XGServerEvent.m
+++ b/Source/x11/XGServerEvent.m
@@ -1220,6 +1220,7 @@ posixFileDescriptor: (NSPosixFileDescriptor*)fileDescriptor
                     xEvent.xfocus.window, cWin->number);
         // Store this for debugging, may not be the real focus window
         generic.currentFocusWindow = cWin->number;
+        generic.desiredFocusWindow = cWin->number;
         if (xEvent.xfocus.serial == generic.focusRequestNumber)
           {
             /*
@@ -1243,18 +1244,24 @@ posixFileDescriptor: (NSPosixFileDescriptor*)fileDescriptor
            * it's not one of ours.
            * If it has gone to our root window - use the icon window.
            * If it has gone to a window - we see if it is one of ours.
+           * If it has gone to our unmapped window - deactivate.
            */
           XGetInputFocus(xEvent.xfocus.display, &fw, &rev);
           NSDebugLLog(@"NSEvent", @"%lu FocusOut\n",
                       xEvent.xfocus.window);
-          if (fw != None && fw != PointerRoot)
+          if (cWin && (fw == cWin->parent) && (cWin->map_state != IsViewable))
+            { // focus switched to WM decorations
+              nswin = GSWindowWithNumber(cWin->number);
+              [self setinputfocus:[[[NSApp mainMenu] window] windowNumber]];
+            }
+          else if (fw != None && fw != PointerRoot)
             {
               generic.cachedWindow = [XGServer _windowForXWindow: fw];
               if (cWin == 0)
                 {
                   generic.cachedWindow = [XGServer _windowForXParent: fw];
                 }
-              if (cWin == 0)
+              if (cWin == 0 || (cWin->map_state != IsViewable))
                 {
                   nswin = nil;
                 }
@@ -1449,7 +1456,7 @@ posixFileDescriptor: (NSPosixFileDescriptor*)fileDescriptor
              * focus, re-do the request.
              */
             if (generic.desiredFocusWindow == cWin->number
-                && generic.focusRequestNumber == 0)
+                && generic.focusRequestNumber == 0 && [NSApp isActive] != NO)
               {
                 NSDebugLLog(@"Focus", @"Refocusing %lu on map notify", 
                             cWin->number);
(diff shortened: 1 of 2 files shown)


fredkiefer:
Sorry, I looked at the code for quite some time. But I am not able to tell whether these changes are an improvement or not. Could somebody please test these changes on different window managers and report back some results?

The WM_TAKE_FOCUS atom ist not WindowMaker specific, it is defined here https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.7

trunkmaster:
The improvement is in order of checks and extra checking.
You are right WM_TAKE_FOCUS is not WindowMaker specific.

Source/x11/XGServerEvent.m
```diff
@@ -1243,18 +1244,24 @@ - (void) processEvent: (XEvent *) event
            * it's not one of ours.
            * If it has gone to our root window - use the icon window.
            * If it has gone to a window - we see if it is one of ours.
+           * If it has gone to our unmapped window - deactivate.
            */
           XGetInputFocus(xEvent.xfocus.display, &fw, &rev);
           NSDebugLLog(@"NSEvent", @"%lu FocusOut\n",
                       xEvent.xfocus.window);
-          if (fw != None && fw != PointerRoot)
+          if (cWin && (fw == cWin->parent) && (cWin->map_state != IsViewable))
```
fredkiefer: Using cwin here is most likely wrong. cwin is a #define on generic.cachedWindow. Which just means the last window that was used for this XGServer. It is a pure coincidence when the focus goes to the parent of the last window for which we did process an event. You seem to expect that the last event was a previous FocusIn event and that cwin is still pointing to that. But this may not be true.

Source/x11/XGServerEvent.m
```diff
@@ -1945,15 +1958,36 @@ - (NSEvent *)_handleTakeFocusAtom: (XEvent)xEvent
...
     }
-  else if (cWin->number == key_num)
+  else if (cWin->ignore_take_focus == YES) // after orderwindow:::
     {
-      NSDebugLLog(@"Focus", @"Reasserting key window");
-      [GSServerForWindow(key_win) setinputfocus: key_num];
+      NSDebugLLog(@"Focus",
+                  @"[TakeFocus] %lu(%li): ignore_take_focus == YES...",
+                  cWin->ident, cWin->number);
+      /* Window was requested to take focus and earlier (orderwindow:::)
```
fredkiefer: You seem to be implying that key_win and cWin are the same window. But why do you think so?

Source/x11/XGServerEvent.m
trunkmaster: This check is for case when the window of application is closing. The next window that receives focus in WindowMaker is window decorations (titlebar, resizebar). So `cWin` at this point contains window with content but in unmapped state. I've added this check to prevent application deactivation and focus loss.
If this is not such situation - fine, `else if` code block executed (old behavior). As I said before I observed focus loss during workspace switch in WindowMaker. 

In any case this code should be harmless in most cases, except the described above.

Source/x11/XGServerEvent.m
trunkmaster: I don't see `key_win` in the cited code? Do you mean code across method? If so, my answer is: I'm not implying this, I'm checking if this is a true at some points. Or you meant some other part of code?

Source/x11/XGServerEvent.m
fredkiefer: But why do you think that cWin is the unmapped content window? It might be in some cases but there is nothing in the code that ensures that.
We should not build our code on coincidences.

trunkmaster: Because these 2 events are going one after other. If it's not - this is not the case I've talked about.

trunkmaster: I need some testing to recall what this change is for.

trunkmaster: While testing I observe situations when this code block was ran. But I can't clearly describe it now. Removed that part of code for now. I'll back to it later maybe.

trunkmaster:
@fredkiefer any othe objections?

trunkmaster:
@fredkiefer ?

Source/x11/XGServerEvent.m
fredkiefer: Just a few lines below 1978, there you check the map_state of keyWin but then set the focus on cWin. In the line 1983 you even get the server from key_win. I really don't understand when this is needed, but the code looks wrong.
Maybe if we could go through the different cases in this method one by one?
The annoying bit about this code here is that all the existing special cases where written because they seemed necessary for a specific window manager. Now you are throwing some of the special stuff away to add your own. The will surely work on your specific WindowMaker setup, but what about other systems?

trunkmaster: `keyWin` and `key_win` represents the same window but different structures: gs_window_device_t and NSWindow respectively. It is a key window as it sees NSApplication. `cWin` is a current focused window (look NSDebugLLog message at line 1922, this is not my text). So code you've pointed me at checks if NSApplication and X11 vision of what the focus window is and performs respective actions. What's wrong with that?

Other window managers are out of my interest. I expect other developers/users try it and give me feedback. Otherwise - no problem, I'm able to maintain this patch in NEXTSPACE.

trunkmaster: You're right actually. I'll check this part of code. Do you have comments to other parts of this patch?

trunkmaster:
Closing this for now.

PR #14, opened 2020-01-12
Correct application icon images in WM

trunkmaster:
If application icon placed into WM_HINTS properties of window (at least for WindowMaker) alpha channel in image missed and icon in Dock looks weird. For example look into [issue #227](https://github.com/trunkmaster/nextspace/issues/227) in NextSpace.

### DIFF
--- a/Source/x11/XGServerWindow.m
+++ b/Source/x11/XGServerWindow.m
@@ -2693,8 +2693,8 @@ static Pixmap xIconMask;
 static BOOL didCreatePixmaps;
 
 Pixmap
-image_mask(Display *xdpy, Drawable draw, const unsigned char *data,
-           int w, int h, int colors, int alpha_treshold)
+alphaMaskForImage(Display *xdpy, Drawable draw, const unsigned char *data,
+                  int w, int h, int colors, int alpha_treshold)
 {
   int j, i;
   unsigned char	ialpha;
@@ -2740,14 +2740,59 @@ image_mask(Display *xdpy, Drawable draw, const unsigned char *data,
   return pix;
 }
 
+// Convert RGBA unpacked to ARGB packed.
+// Packed ARGB values are layed out as ARGB on big endian systems
+// and as BGRA on low endian systems
+void
+swapColors(unsigned char *image_data, int width, int height,
+           int samples_per_pixel, int bytes_per_row)
+{
+  NSInteger     x, y;
+  unsigned char *data;
+  unsigned char *r, *g, *b, *a;
+
+  data = image_data;
+  r = data;
+  g = data + 1;
+  b = data + 2;
+  a = data + 3;
+  for (y = 0; y < height; y++)
+    {
+      unsigned char *d = data;
+      for (x = 0; x < width; x++)
+	{
+#if GS_WORDS_BIGENDIAN
+          // RGBA -> ARGB
+          unsigned char _d = d[3];
+          *r = _d;
+          *g = d[0];
+          *b = d[1];
+          *a = d[2];
+#else
+          // RGBA -> BGRA
+          unsigned char _d = d[0];
+          *r = d[2];
+          // *g = d[1];
+          *b = _d;
+          // *a = d[3];
+#endif
+          r += 4;
+          g += 4;
+          b += 4;
+          a += 4;
+	  d += samples_per_pixel;
+	}
+      data += bytes_per_row;
+    }
+}
+
+
 - (int) _createAppIconPixmaps
 {
   NSBitmapImageRep *rep;
-  int              i, j, width, height, samples, screen;
-  unsigned char    *data;
+  int              width, height, colors, screen;
   RContext         *rcontext;
   RXImage          *rxImage;
-  char             *r, *g, *b, *a;
 
   NSAssert(!didCreatePixmaps, @"called _createAppIconPixmap twice");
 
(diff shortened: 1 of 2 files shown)


Source/x11/XGServerWindow.m
```diff
@@ -3072,10 +3104,11 @@ - (void) restrictWindow: (int)win toImage: (NSImage*)image
...
 	      {
-		pixmap = xgps_cursor_mask(dpy, GET_XDRAWABLE(window),
-					  [rep bitmapData],
-					  [rep pixelsWide], [rep pixelsHigh],
-					  [rep samplesPerPixel]);
+		pixmap = image_mask(dpy, GET_XDRAWABLE(window),
+                                    [rep bitmapData],
+                                    [rep pixelsWide], [rep pixelsHigh],
+                                    [rep samplesPerPixel],
+                                    ALPHA_THRESHOLD);
```
fredkiefer: What is the value of ALPHA_THRESHOLD at this point?

Source/x11/XGServerWindow.m
```diff
@@ -2707,62 +2692,109 @@ - (void) setWindowdevice: (int)win forContext: (NSGraphicsContext *)ctxt
...
-  h = [rep pixelsHigh];
-  w = [rep pixelsWide];
-  samples = [rep samplesPerPixel];
   rcontext = [self xrContextForScreen: screen];
+  width = [rep pixelsWide];
+  height = [rep pixelsHigh];
+  samples = [rep samplesPerPixel];
 
-  for (i = 0; i < h; i++)
+  /**/
```
fredkiefer: What is this empty comment for?

Source/x11/XGServerWindow.m
```diff
@@ -2707,62 +2692,109 @@ - (void) setWindowdevice: (int)win forContext: (NSGraphicsContext *)ctxt
...
+  rxImage = RCreateXImage(rcontext, rcontext->depth, width, height);
+  xIconPixmap = XCreatePixmap(dpy, rcontext->drawable,
```
fredkiefer: This line could now be moved further down to where xIconPixmap gets used.

Source/x11/XGServerWindow.m
```diff
@@ -2707,62 +2692,109 @@ - (void) setWindowdevice: (int)win forContext: (NSGraphicsContext *)ctxt
...
     {
       unsigned char *d = data;
-      for (j = 0; j < w; j++)
+      for (j = 0; j < width; j++)
 	{
-	  pixelRColor.red = d[0];
-	  pixelRColor.green = d[1];
-	  pixelRColor.blue = d[2];
-
-	  RGetClosestXColor(rcontext, &pixelRColor, &pixelColor);
```
fredkiefer: With this change the code won't run that well in indexed displays but I am not sure whether any GNUstep application was executed in such an environment in the last twenty years, so for me this change is fine.

Source/x11/XGServerWindow.m
```diff
@@ -2707,62 +2692,109 @@ - (void) setWindowdevice: (int)win forContext: (NSGraphicsContext *)ctxt
...
-	  XSetForeground(dpy, pixgc, pixelColor. pixel);
-	  XDrawPoint(dpy, xIconPixmap, pixgc, j, i);
+          *r = d[2];
```
fredkiefer: There should already be code in the x11 backend where we switch between rgba and bgra, could this be reused here?

Source/x11/XGServerWindow.m
```diff
@@ -2707,62 +2692,109 @@ - (void) setWindowdevice: (int)win forContext: (NSGraphicsContext *)ctxt
...
-  XFreeGC(dpy, pixgc);
-
-  xIconMask = xgps_cursor_mask(dpy, ROOT, [rep bitmapData],
-			       [rep pixelsWide],
-			       [rep pixelsHigh],
-			       [rep samplesPerPixel]);
+  XPutImage(dpy, xIconPixmap, rcontext->copy_gc, rxImage->image,
+            0, 0, 0, 0, width, height);
+  RDestroyXImage(rcontext, rxImage);
+  /**/
```
fredkiefer: Another empty comment

Source/x11/XGServerWindow.m
trunkmaster: #define is at line 3080.

Source/x11/XGServerWindow.m
trunkmaster: Done.

Source/x11/XGServerWindow.m
trunkmaster: I've created new function swapColors() and use in 2 places.

Source/x11/XGServerWindow.m
```diff
@@ -2707,62 +2692,134 @@ - (void) setWindowdevice: (int)win forContext: (NSGraphicsContext *)ctxt
...
+
+  pix = XCreatePixmapFromBitmapData(xdpy, draw, (char *)aData, w, h,
+				    1L, 0L, 1);
+  free(aData);
+  return pix;
+}
+
+// Convert RGBA unpacked to ARGB packed.
+// Packed ARGB values are layed out as ARGB on big endian systems
+// and as BGRA on low endian systems
```
fredkiefer: "little" endian

Source/x11/XGServerWindow.m
```diff
@@ -4017,55 +4075,9 @@ - (void) _DPSsetcursor: (Cursor)c : (BOOL)set
...
-	  *cData++ = 0xff;
-	}
-    }
-
-  pix = XCreatePixmapFromBitmapData(xdpy, draw, (char *)aData, w, h,
-				    1L, 0L, 1);
-  free(aData);
-  return pix;
-}
+#define ALPHA_THRESHOLD 158
```
fredkiefer: I really would prefer to move this define before the first usage of the value. This may work nowadays but I am still used to the old way :-(

trunkmaster: Actually this is a duplicate. The other is at line 3105. Removed.

Source/x11/XGServerWindow.m
trunkmaster: Fixed, thanks.

PR #15, opened 2020-01-15
TakeFocus request handling fix

trunkmaster:
This fix primary goal is to manage the following case:
- user opens application on workspace0 with some window (say Terminal);
- user switched to workspace1;
- user activates Terminal by double-clicking on appicon;
- Terminal main menu will be shown, but input focus was left on Terminal window left on workspace0.

Also, I think I've made _handleTakeFocusAtom:forContext: more correct and maintainable due to the usage of window that receives TakeFocus request instead of last window used by application.

### DIFF
--- a/Source/x11/XGServerEvent.m
+++ b/Source/x11/XGServerEvent.m
@@ -1914,18 +1914,16 @@ posixFileDescriptor: (NSPosixFileDescriptor*)fileDescriptor
   int key_num;
   NSWindow *keyWindow;
   NSEvent *e = nil;
-  gswindow_device_t *this_window;
   gswindow_device_t *key_window;
   
   keyWindow = [NSApp keyWindow];
   key_num = [keyWindow windowNumber];
   key_window = [XGServer _windowWithTag:key_num];
-  this_window = [XGServer _windowForXWindow:xEvent.xfocus.window];
   
   NSDebugLLog(@"Focus",
-              @"TakeFocus received by: %li (%lu) (focused = %lu, key = %d, cWin = %lu)",
-              this_window->number, xEvent.xfocus.window,
-              generic.currentFocusWindow, key_num, cWin->number);
+              @"TakeFocus received by: %li (%lu) (focused = %lu, key = %d)",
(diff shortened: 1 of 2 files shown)


Source/x11/XGServerEvent.m
```diff
@@ -1912,27 +1912,42 @@ - (NSEvent *)_handleTakeFocusAtom: (XEvent)xEvent
...
-   */
-  if (key_num == cWin->number)
-    cWin->ignore_take_focus = NO;
+  gswindow_device_t *this_window;
+  gswindow_device_t *key_window;
   
+  keyWindow = [NSApp keyWindow];
+  key_num = [keyWindow windowNumber];
+  key_window = [XGServer _windowWithTag:key_num];
+  this_window = [XGServer _windowForXWindow:xEvent.xfocus.window];
```
fredkiefer: Using cwin instead of this_window is better as this is is already coached for all ClientMessage events.
In the following log message the tow should always produce the same number. I would suggest to stick with cwin as this reduces the size of the change and keeps the code closer to the surrounding.

trunkmaster: Thanks, done.

Source/x11/XGServerEvent.m
```diff
@@ -1912,27 +1912,40 @@ - (NSEvent *)_handleTakeFocusAtom: (XEvent)xEvent
...
-   * e.g. when ordering out a window with focus then ordering in the key window.   
-   * it might search for a window until one accepts its take focus request.
-   */
-  if (key_num == cWin->number)
-    cWin->ignore_take_focus = NO;
+  gswindow_device_t *key_window;
+  
+  keyWindow = [NSApp keyWindow];
+  key_num = [keyWindow windowNumber];
+  key_window = [XGServer _windowWithTag:key_num];
```
fredkiefer: Could we move this line to 1974 as that structure only gets used there?

trunkmaster: No problem. Done.

PR #17, opened 2020-01-26
Wayland

ivucica:
This includes contributions from Sergio L. Pascual and Ladislav Michl @3x380V. Both have performed copyright assignment to FSF.

This likely requires a rebase+merge of https://github.com/ivucica/libs-gui/tree/ivucica-wayland as well.

This currently does not run on my machine, as I'm hitting:

```
    if (!wlconfig->compositor || !wlconfig->wm_base) {
        [NSException raise: NSWindowServerCommunicationException
                    format: @"Unable to get compositor"];
```

Opening this PR for easier viewing anyway.


Headers/cairo/WaylandCairoSurface.h
```diff
@@ -0,0 +1,16 @@
+/*
+   Copyright (C) 2016 Sergio L. Pascual <slp@sinrega.org>
```
fredkiefer: Could we change all the copyright messages here and add the standard GNUstep header?

Source/cairo/CairoContext.m
```diff
@@ -70,6 +70,12 @@
 #    define _CAIRO_SURFACE_CLASSNAME Win32CairoSurface
 #    include "cairo/Win32CairoSurface.h"
 #  endif /* USE_GLITZ */
+#elif BUILD_SERVER == SERVER_wayland
+#  include "wayland/WaylandServer.h"
+#  define _CAIRO_GSTATE_CLASSNAME CairoGState
+#  include "cairo/CairoGState.h"
```
fredkiefer: i would move the includes before the defines.

Source/cairo/WaylandCairoSurface.m
```diff
@@ -0,0 +1,204 @@
...
+    wl_surface_attach(window->surface, window->buffer, 0, 0);
+    window->wcs = self;
+
+    return self;
+}
+
+- (void) dealloc
+{
+    struct window *window = (struct window*) gsDevice;
+    NSDebugLog(@"WaylandCairoSurface: dealloc win=%d", window->window_id);
```
fredkiefer: Shouldn't this free the shm buffer?

Source/cairo/WaylandCairoSurface.m
```diff
@@ -0,0 +1,204 @@
...
+	NSWarnMLog(@"cairo initial window error status: %s\n",
+		   cairo_status_to_string(cairo_surface_status(_surface)));
+    }
+
+    cairo_surface_get_device_offset(cairo_surface, &backupOffsetX, &backupOffsetY);
+    cairo_surface_set_device_offset(cairo_surface, 0, 0);
+
+    cairo_t *cr = cairo_create(cairo_surface);
+    if (width != window->width && 0) {
+	NSDebugLog(@"fake drawing");
```
fredkiefer: What is this branch for?

Source/cairo/WaylandCairoSurface.m
```diff
@@ -0,0 +1,204 @@
...
+	cairo_set_font_size(cr, 12);
+	cairo_set_source_rgba(cr, 1.0, 1.0, 1.0, 1.0);
+	cairo_move_to(cr, 30, 30);
+	cairo_show_text(cr, "Hello, world!");
+    } else {
+	NSDebugLog(@"real drawing");
+
+	cairo_rectangle(cr, x, y, width, height);
+	cairo_clip(cr);
+	cairo_set_source_surface(cr, cairo_surface, 0, 0);
```
fredkiefer: Doesn't this copy from cairo_surface to cairo_surface?

Source/wayland/GNUmakefile
```diff
@@ -0,0 +1,50 @@
+#
+#  Main makefile for GNUstep Backend win32
+#
+#  Copyright (C) 2002 Free Software Foundation, Inc.
```
fredkiefer: We probably should change the year and the name.

Source/wayland/GNUmakefile.preamble
```diff
@@ -0,0 +1,51 @@
+
+#  GNUmakefile.preamble
+#
+#  Copyright (C) 2002 Free Software Foundation, Inc.
```
fredkiefer: Same here

Source/wayland/WaylandServer.m
```diff
@@ -0,0 +1,1482 @@
...
+    }
+
+    WaylandConfig *wlconfig = data;
+    struct window *window = wl_surface_get_user_data(surface);
+    float sx = wl_fixed_to_double(sx_w);
+    float sy = wl_fixed_to_double(sy_w);
+
+    wlconfig->pointer.x = sx;
+    wlconfig->pointer.y = sy;
+    wlconfig->pointer.focus = window;
```
fredkiefer: Shouldn't this send an event to the frontend?

Source/wayland/WaylandServer.m
```diff
@@ -0,0 +1,1482 @@
...
+	    wlconfig->pointer.last_click_y = wlconfig->pointer.y;
+	}
+
+	switch (button) {
+	case BTN_LEFT:
+	    eventType = NSLeftMouseDown;
+	    break;
+	case BTN_RIGHT:
+	    eventType = NSRightMouseDown;
+	    break;
```
fredkiefer: What happens for all these other buttons on my mouse?

Source/wayland/WaylandServer.m
```diff
@@ -0,0 +1,1482 @@
...
+				 deltaZ: 0.];
+
+    [GSCurrentServer() postEvent: event atStart: NO];
+}
+
+static void
+pointer_handle_axis(void *data, struct wl_pointer *pointer,
+		    uint32_t time, uint32_t axis, wl_fixed_t value)
+{
+}
```
fredkiefer: This should send scroll wheel events.

Source/wayland/WaylandServer.m
```diff
@@ -0,0 +1,1482 @@
...
+    window->instance = self;
+    window->is_out = 0;
+    window->width = width = NSWidth(frame);
+    window->height = height = NSHeight(frame);
+    window->pos_x = frame.origin.x;
+    window->pos_y = NSToWayland(window, frame.origin.y);
+    window->window_id = wlconfig->last_window_id;
+    window->surface = wl_compositor_create_surface(wlconfig->compositor);
+    if (!window->surface) {
+	NSDebugLog(@"can't create wayland surface");
```
fredkiefer: Shouldn't we free the window here?

Source/wayland/WaylandServer.m
```diff
@@ -0,0 +1,1482 @@
...
+    return 0;
+}
+
+/** Backends can override this method to return an array of window numbers
+    ordered front to back.  The front most window being the first object
+    in the array.
+    The default implementation returns the visible windows in an
+    unspecified order.
+ */
+- (NSArray *) windowlist
```
fredkiefer: Why not just use the default implementation?

Source/wayland/WaylandServer.m
```diff
@@ -0,0 +1,1482 @@
...
+}
+
+- (void) setShadow: (BOOL)hasShadow : (int)win
+{
+    NSDebugLog(@"setshadow");
+}
+
+- (NSPoint) mouselocation
+{
+    NSDebugLog(@"mouselocation");
```
fredkiefer: Why not call the following method here?

Source/wayland/WaylandServer.m
```diff
@@ -0,0 +1,1482 @@
...
+	y = wlconfig->pointer.captured_y;
+	} else*/ {
+	NSDebugLog(@"NOT captured");
+	x = wlconfig->pointer.x;
+	y = wlconfig->pointer.y;
+
+	if (window) {
+	    x += window->pos_x;
+	    y += window->pos_y;
+	    win = &window->window_id;
```
fredkiefer: Check for win being NULL is missing.

ivucica:
I'll continue work on this after the next release.

Headers/cairo/WaylandCairoSurface.h
ivucica: Done. We'll need a review from @slp -- I think copyright assignment happened so that the copyright ownership transfer was intended, but let's confirm this.

Source/cairo/CairoContext.m
ivucica: Done. FWIW this is not the case in other `#if`s.

Source/wayland/GNUmakefile
ivucica: Also did `s/win32/wayland`, and added Sergio and Ladislav.

Source/wayland/GNUmakefile.preamble
ivucica: Done. I added Sergio as well.

Source/wayland/WaylandServer.m
ivucica: Done

Source/wayland/WaylandServer.m
ivucica: Interestingly, it was not even writing into `*win`...

ivucica:
Sorry for the noise and presumably numerous emails. Github review UI was surprisingly not letting me comment directly in the pull request after I started a review; presumably I was supposed to write replies 'somewhere else'.

I think your comments are useful, so I am reluctant to just go ahead and merge as-is, thus losing the information in this then-merged PR.

I intended to run the code today, but I only got around to rebasing the code, and addressing the most superficial of problems you spotted. Thank you for looking at this in-depth; applying some fixes will be tricky as well, as I am new to Wayland and this new code. Perhaps soon.

ivucica:
@slp It looks like I can't directly assign you this review (presumably because you were not added to the GH org, not sure).

However, your input is required for the changes to the headers. Has the copyright been assigned and may I change the headers as visible in my commits?

Headers/cairo/WaylandCairoSurface.h
slp: @ivucica Sure! Please change that to the standard GNUstep header.

ivucica:
In https://github.com/gnustep/libs-back/pull/17/commits/41b2d294ed7393a5331c3aefa205fad41fabd502 I managed to get stuff to paint by not committing the surface unless we first received a configure request and have acked it. Not doing this is against the protocol, and Weston actively drops the connection if we do this.

Only updating one location for `wl_surface_commit()`, the one in `WaylandCairoSurface()`, does the trick; in the other location, apparently we are supposed to commit this surface, even if unconfigured.

We still get a lot of garbage, and behavior is generally very broken when it comes to input and clicking around.

Here's SystemPreferences.app:

image

Here's Calculator.app:

image

ivucica:
Some resources for understanding that commit:

- https://github.com/wayland-project/weston/blob/24c0f83778f3db8476cd875580876d1d1a27a13a/libweston-desktop/xdg-shell.c#L1250
- https://patches.videolan.org/patch/20875/
- https://github.com/michaelforney/st/pull/8#issuecomment-403615287 (albeit to a lesser extent)

Source/cairo/WaylandCairoSurface.m
ivucica: Attempting to find out exactly what needs to be freed is proving to be tricky.

`munmap()`ing the memory seemingly used by the Cairo surface without freeing the Cairo surface seems wrong to me, and calling `cairo_surface_destroy()` here to decrease the refcount and likely destroy the surface seems to be causing crash within `libwayland-client` code.

I'm therefore leaving an extensive comment here and marking this resolved.

Source/cairo/WaylandCairoSurface.m
ivucica: I can't tell, and it was disabled in the patches we received, so I removed it. Presumably stub drawing code left-over from the original author.

Source/cairo/WaylandCairoSurface.m
ivucica: It does seem to be doing so. I cannot immediately decipher what's the buffer it's supposed to copy from, so in the interest of merging this code, I'm leaving a FIXME.

Source/wayland/WaylandServer.m
ivucica: I have found the relevant constants and added a TODO to send `NSMouseEntered` and `NSMouseExited`. However, I will not implement this at this time, as no other backend (!) is even referencing these constants.

Source/wayland/WaylandServer.m
ivucica: The third constants I found supported is `NSOtherMouse{Down,Up}`, so I added support for that.

If other buttons on your mouse are supported by libinput, I don't immediately know how to map them to AppKit constants: `BTN_SIDE`, `BTN_EXTRA`, `BTN_FORWARD`, `BTN_BACK` and any other ones. :-)

Source/wayland/WaylandServer.m
ivucica: Done, albeit it feels clumsily. Then again, the X11 implementation is not that beautiful either...

Source/wayland/WaylandServer.m
ivucica: I'll assume original author took `GSDisplayServer` as the base implementation and then implemented individual methods as required. Deleted the useless override.

Source/wayland/WaylandServer.m
ivucica: Done.

ivucica:
@fredkiefer Would you like to take a final look at this before we merge this? As we agreed, we should merge it in its incomplete and broken state.

If you LGTM this, I'll rebase as necessary and push to the main branch, updating any dates as required. Let me know.

PR #18, opened 2020-01-26
Dynamic screen resolution change support

trunkmaster:
The goal of forthcoming set of changes is provide better multi-monitor support to GNUstep.
This part brings XRandR extension support to:
- receive and handle XRandR screen resolution change events;
- send NSApplicationDidChangeScreenParameters notification;
- retreive and return screen dimensions from XRandR structures (boundsForRect) instead of Xorg display's;
- use `boundsForRect` instead of DisplayHeight() - it returns old height of screen after screen pixel dimension changes.

The next important change is update origin of windows in new screen dimensions - libs-gui [PR #50](https://github.com/gnustep/libs-gui/pull/50).


Headers/x11/XGServer.h
```diff
@@ -60,6 +60,7 @@ typedef enum {
   Window	    grabWindow;
   struct XGGeneric  generic;
   id                inputServer;
+  int               randrEventBase, randrErrorBase;
```
fredkiefer: I personally prefer to have (instance) variables defined on separate lines.

Source/x11/XGDragView.m
```diff
@@ -54,7 +54,7 @@
 
 #define	DRAGWINDEV [XGServer _windowWithTag: [_window windowNumber]]
 #define	XX(P)	(P.x)
-#define	XY(P)	(DisplayHeight(XDPY, DRAGWINDEV->screen) - P.y)
+#define	XY(P)	([GSCurrentServer() boundsForScreen: DRAGWINDEV->screen].size.height - P.y)
```
fredkiefer: Getting a field from a structure returned by an ObjC call is a dangerous operation. On one systems null may be returned instead of a valid structure. We need to check whether the the recipient is not nil. But in this specific case we are hopefully save.

Source/x11/XGServerEvent.m
```diff
@@ -1894,6 +1898,25 @@ - (void) processEvent: (XEvent *) event
...
             break;
           }
+#endif
+#ifdef HAVE_XRANDR
+        int randr_event_type = randrEventBase + RRScreenChangeNotify;
+        if (xEvent.type == randr_event_type
+            && (xEvent.xconfigure.window == RootWindow(dpy, defScreen)))
+          {
+            // Check if other RandR events are waiting in the queue.
+            XSync(dpy,0);
```
fredkiefer: Space after the comma.

Source/x11/XGServerEvent.m
```diff
@@ -1894,6 +1898,25 @@ - (void) processEvent: (XEvent *) event
...
+            while (XCheckTypedEvent(dpy, randr_event_type, &xEvent)) {;}
+                    
+            XRRUpdateConfiguration(event);
+            // Regenerate NSScreens
+            [NSScreen resetScreens];
+            // Notify application about screen parameters change
```
fredkiefer: Should we move this notification into the NSScreen method?

Source/x11/XGServerWindow.m
```diff
@@ -4509,13 +4512,33 @@ - (NSSize) resolutionForScreen: (int)screen_num
...
 - (NSRect) boundsForScreen: (int)screen
 {
- if (screen < 0 || screen >= ScreenCount(dpy))
-   {
-     NSLog(@"Invalidparam: no screen %d", screen);
-     return NSZeroRect;
-   }
- return NSMakeRect(0, 0, DisplayWidth(dpy, screen),
-		   DisplayHeight(dpy, screen));
+  NSRect boundsRect = {{0,0},{0,0}};
```
fredkiefer: I think this is NSZeroRect, why not use it here as well?

Source/x11/XGServerWindow.m
```diff
@@ -4509,13 +4512,33 @@ - (NSSize) resolutionForScreen: (int)screen_num
...
+  
+  if (screen < 0 || screen >= ScreenCount(dpy))
+    {
+      NSLog(@"Invalidparam: no screen %d", screen);
+      return NSZeroRect;
+    }
+  
+#ifdef HAVE_XRANDR
+  XRRScreenResources *screen_res;
```
fredkiefer: As this operations look expensive, should we cache this information in NSScreen resetScreens and reuse the cached values?

Headers/x11/XGServer.h
trunkmaster: No problem. Done.

Source/x11/XGServerWindow.m
trunkmaster: Interesting idea, thanks. I'll do it.

Source/x11/XGServerEvent.m
trunkmaster: `resetScreens` is a GNUstep addition. I'm not sure about exact purpose of this method. Anyone can call it without actual screen change. Do you think it's safe?

Source/x11/XGServerEvent.m
trunkmaster: Done.

Source/x11/XGServerWindow.m
trunkmaster: Thanks, done.

trunkmaster:
Actually current screen handling implementation is limited. NSScreen percieves screen as a display/monitor device. Other side - backend - manipulates X11 screen - the sum of displays bounds (overlapped or placed side by side). Saying that, I plan to change this part of -back -gui interconnection.
Meanwhile I merge this PR and start making enhancements as new PR.

Source/x11/XGServerEvent.m
fredkiefer: Why would code call that method without screen changes? What we should try to figure out is where that notification code belongs logically. If it is more likely that gui objects are listening for that notification I would prefer to have it there. That way you understand what is going on by just looking at the gui code. You won't have to dive into back to see the logic.

trunkmaster: > Why would code call that method without screen changes? 

For example when NSScreen or some application want to refresh screens information.

> What we should try to figure out is where that notification code belongs logically. 

I think this notification is at the right place - where screen changes (RandR) event processed.

> If it is more likely that gui objects are listening for that notification I would prefer to have it there. That way you understand what is going on by just looking at the gui code. You won't have to dive into back to see the logic.

I understand. But you have to dive into back code anyway because `DESTROY(screenArray)` line in `resetScreens` will not clear the situation up. You probably need to find out who have called `resetScreens`. From arhitectural point of view `resetScreens` can be replaced with some other method (screenArray readwrite property) because `resetScreens` has one purpose - force NSScreen to update its information about displays on next call to `screens`.

PR #19, opened 2020-01-29
Screens handling with RandR fixes

trunkmaster:
In addition to the previous PR I've made some optiomizations and fixes:
- implemented caching of monitors parameters;
- fixed monitors enumeration - NSScreen receives correct array with indices to monitors parameters cache;
- fixed/unified handling count of monitors/screens with RandR and without it.
- boundsForScreen now should run much faster because of cached parameters use.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,4 @@
-2020-01-26  Sergii Stoian  <stoyan255@gmail.com>
+2020-01-30  Sergii Stoian  <stoyan255@gmail.com>
 
 	* Headers/x11/XGServer.h: new structure `MonitorDevice` and ivar
 	`monitors`for holding cache of monitor's parameters (depth, resolution,
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,74 @@
+2020-02-03  Sergii Stoian  <stoyan255@gmail.com>
+
+	* Source/x11/XWindowBuffer.m (windowBufferForWindow:depthInfo:): use
+	renamed	XGServer's `screenVisual` and `screenDepth` methods.
+
+	* Source/x11/XIMInputServer.m (initWithDelegate:name:):
+	use `+xDisplay` instead	of `xCurrentDisplay`. 
+	(clientWindowRect:): ditto.
+
+	* Headers/x11/XGServerWindow.h (_gswindow_device_t):
+	renamed `screen` to `screen_id`. `monitor_id` element was added -
+	this is an index of element in MonitorDevice array.
+	* Source/x11/XGServerWindow.m:
+	Removed `screen` parameter from methods which contains it in name.
+	Methods were renamed to remove `Screen` and `FromScreen`.
+	(_OSFrameToXFrame:for:): use `monitor_id` instead of `screen`.
+	(_OSFrameToXHints:for:): ditto.
+	(_XFrameToOSFrame:for:): ditto.
+	(_checkStyle:): use renamed methods and `screen_id` window structure
+	element.
+	(_rootWindow): renamed from _rootWindowForScreen:, use `defScreen`
+	instead of removed parameter `screen`, set window structure element
+	`monitor_id` to 0.
+	(_createBuffer:): use `screen_id` window structure.
+	(window::::): use renamed methods, `defScreen`, `screen_id` and
+	`monitor_id` window scructure elements.
+	(nativeWindow:::::): ditto.
+	(setbackgroundcolor::): ditto.
+	(miniwindow:): ditto.
+	(_createAppIconPixmaps): removed unused `screen` local variable, use
+	renamed `screenRContext` method.
+	(orderwindow:::): use `screen_id` window structure element.
+	(movewindow::): use `screen_id` window structure element as a parameter
+	to `boundsForScreen`.
+	(windowbounds:): ditto.
+	(windowlist): use renamed `_rootWindow` method.
+	(setMouseLocation:onScreen:): user renamed methods.
+	(_blankCursor): ditto.
+	(imagecursor:::): ditto
+	(recolorcursor:::): ditto.
+	(screenList): change with assumption that Xlib screen is one per application.
+	Failback code to RandR-related always creates arrays with one element (no
+	enumeration needed). Create `monitor` structure in failback (non-RandR) code.
+	Use `screen_id` MonitorDevice structure element to hold	Xlib screen ID.
+	(windowDepthForScreen:): assume that `screen` parameter is a Xlib screen ID,
+	simplify code - `screen` will be used with RandR or not.
+	(availableDepthsForScreen:): changed with assumption that `screen` is a index
+	of element in `monitors` structure.
+
+	* Source/x11/XGServerEvent.m:
+	use renamed methods and `screen_id` window structure element.
+
+	* Headers/x11/XGServer.h
+	Removed `screen` parameter from methods which contains it in name.
+	Methods were renamed to remove `Screen` and `FromScreen`.
+	(MonitorDevice): `screen_id` was added.
+
+	* Source/x11/XGServer.m:
+	use `screen_id` and rename mathods according to XGServer,h.
+	(_initXContext): initialize `monitors` array early here before first
+	use (calls to XGServerWindow methods).
+
+	* Source/x11/XGDragView.m (XDPY): use `monitor_id` in `boundForScreen:`
+	call.
+
+	* Source/cairo/XGCairoSurface.m: use `xDisplayRootWindow` XGServer's method.
+
+	* Source/art/ARTContext.m (setupDrawInfo:): use `xDisplayRootWindow`
+	XGServer's method.
+
+
 2020-01-31  Sergii Stoian  <stoyan255@gmail.com>
 
 	* Source/x11/XGServerWindow.m (screenList): join RandR and non-RandR
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,6 +9,14 @@
 	indices to get access to monitors` items.
 	(boundsForScreen:): use cached `frame` parameter of specified monitor.
 	Use `monitorsCount` instead of ScreenCount().
+	(windowDepthForScreen:): renamed method parameter; append `x_` prefix to
+	`Screen *` internal variable; validate `screen` parameter value; use
+	`defScreen` for RandR enabled code.
+	(availableDepthsForScreen:): ditto.
+	(resolutionForScreen:): renamed method parameter; validate `screen`
+	parameter value.
+	(boundsForScreen:): renamed method parameter.
+
 	* Source/x11/XGServer.m (dealloc): free `monitors` if it was used.
 
 2020-01-26  Sergii Stoian  <stoyan255@gmail.com>
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2020-01-31  Sergii Stoian  <stoyan255@gmail.com>
+
+	* Source/x11/XGServerWindow.m (screenList): join RandR and non-RandR
+	code. Failback to Xlib generic method if any no RandR available or
+	RandR function call was unsuccessful. In Xlib generic method fill in
+	monitors array with values.
+	(boundsForScreen:): use values from cached `monitors` array since
+	-screenList fills it on both cases with or without RandR available.
+
 2020-01-30  Sergii Stoian  <stoyan255@gmail.com>
 
 	* Headers/x11/XGServer.h: new structure `MonitorDevice` and ivar
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,16 @@
+2020-02-07  Sergii Stoian  <stoyan255@gmail.com>
+
+	* Source/x11/XGServerWindow.m (_XFrameToOSFrame:for:): use Xlib
+	screen height instead of monitor's because we convert coordinates
+	from Xlib area.
+	(movewindow::): use Xlib screen height because we're operating in
+	Xlib coordinate system.
+	(screenList): fill monitor depth with value returned by
+	windowPathForScreen with monitor at index 0 (screen_id already set so
+	method will get correct screen_id).
+
+	* Source/x11/XGServerEvent.m (processEvent:): fixed indentaion.
+
 2020-02-07 Fred Kiefer <FredKiefer@gmx.de>
 
 	* Source/x11/XGServerWindow.m (swapColors): Made code more
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,18 @@
 2020-01-26  Sergii Stoian  <stoyan255@gmail.com>
 
+	* Headers/x11/XGServer.h: new structure `MonitorDevice` and ivar
+	`monitors`for holding cache of monitor's parameters (depth, resolution,
+	frame). New ivar `monitorsCount` - holds number of items in `monitors` for
+	RandR or X11 screen count otherwise.
+	* Source/x11/XGServerWindow.m (screenList): new method for RandR mode -
+	enumerates monitors and caches their parameters. Returns array of monitors'
+	indices to get access to monitors` items.
+	(boundsForScreen:): use cached `frame` parameter of specified monitor.
+	Use `monitorsCount` instead of ScreenCount().
+	* Source/x11/XGServer.m (dealloc): free `monitors` if it was used.
+
+2020-01-26  Sergii Stoian  <stoyan255@gmail.com>
+
 	* Source/x11/XGServerWindow.m (boundsForScreen:): use `screen` variable
 	to identify output in RandR screen resources. Use `boundsRect` local
 	variable as return vaalue storage. Cleanup.
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,15 @@
+2020-02-07  Sergii Stoian  <stoyan255@gmail.com>
+
+	* Source/x11/XGServerWindow.m (_OSFrameToXFrame:for:):,
+	(_OSFrameToXHints:for:): use Xlib screen height instead of monitor's
+	because we convert coordinates to Xlib area.
+
+	* Source/x11/XGServerEvent.m (processEvent:): update monitor_id from
+	NSWindow's screen number (it could change after event processing).
+	(mouseLocationOnScreen:window:): use Xlib screen height instead of
+	monitor's. Added comment to code that should be probably removed
+	later.
+
 2020-02-03  Sergii Stoian  <stoyan255@gmail.com>
 
 	* Source/x11/XWindowBuffer.m (windowBufferForWindow:depthInfo:): use
--- a/Headers/x11/XGServer.h
+++ b/Headers/x11/XGServer.h
@@ -52,6 +52,12 @@ typedef enum {
   XGDM_PORTABLE
 } XGDrawMechanism;
 
+typedef struct MonitorDevice {
+  NSWindowDepth depth;
+  NSSize resolution;
+  NSRect frame;
+} MonitorDevice;
+
 @interface XGServer : GSDisplayServer
 {
   Display           *dpy;
@@ -62,6 +68,7 @@ typedef enum {
   id                inputServer;
   int               randrEventBase;
   int               randrErrorBase;
+  MonitorDevice     *monitors;
 }
 
 + (Display*) currentXDisplay;
--- a/Headers/x11/XGServer.h
+++ b/Headers/x11/XGServer.h
@@ -53,6 +53,7 @@ typedef enum {
 } XGDrawMechanism;
 
 typedef struct MonitorDevice {
+  int screen_id;
   NSWindowDepth depth;
   NSSize resolution;
   NSRect frame;
@@ -72,19 +73,19 @@ typedef struct MonitorDevice {
   unsigned          monitorsCount;
 }
 
-+ (Display*) currentXDisplay;
-- (Display*) xDisplay;
++ (Display *) xDisplay;
+- (Display *) xDisplay;
+- (Window) xDisplayRootWindow;
 - (Window) xAppRootWindow;
 
-- (void *) xrContextForScreen: (int)screen_number;
-- (Visual *) visualForScreen: (int)screen_number;
-- (int) depthForScreen: (int)screen_number;
+- (void *) screenRContext;
+- (Visual *) screenVisual;
+- (int) screenDepth;
+- (XGDrawMechanism) screenDrawMechanism;
 
-- (XGDrawMechanism) drawMechanismForScreen: (int)screen_number;
 - (void) getForScreen: (int)screen_number pixelFormat: (int *)bpp_number 
                 masks: (int *)red_mask : (int *)green_mask : (int *)blue_mask;
-- (Window) xDisplayRootWindowForScreen: (int)screen_number;
-- (XColor) xColorFromColor: (XColor)color forScreen: (int)screen_number;
+- (XColor) xColorFromColor: (XColor)color;
 
 + (void) waitAllContexts;
 @end
--- a/Headers/x11/XGServer.h
+++ b/Headers/x11/XGServer.h
@@ -69,6 +69,7 @@ typedef struct MonitorDevice {
   int               randrEventBase;
   int               randrErrorBase;
   MonitorDevice     *monitors;
+  unsigned          monitorsCount;
 }
 
 + (Display*) currentXDisplay;
--- a/Headers/x11/XGServerWindow.h
+++ b/Headers/x11/XGServerWindow.h
@@ -84,8 +84,9 @@ typedef struct _gswindow_device_t {
   Display               *display;      /* Display this window is on */
   Window                ident;         /* Window handle */
   Window                root;          /* Handle of root window */
-  Window		parent;        /* Handle of parent window */
-  int                   screen;        /* Screeen this window is on */
+  Window                parent;        /* Handle of parent window */
+  int                   screen_id;     /* Screeen this window is on */
+  int                   monitor_id;    /* Physical monitor this window is on */
   GC                    gc;            /* GC for drawing */
   long                  number;        /* Globally unique identifier */
   unsigned int          depth;         /* Window depth */
--- a/Source/x11/XGServer.m
+++ b/Source/x11/XGServer.m
@@ -503,6 +503,10 @@ _parse_display_name(NSString *name, int *dn, int *sn)
   DESTROY(inputServer);
   [self _destroyServerWindows];
   NSFreeMapTable(screenList);
+  if (monitors != NULL)
+    {
+      NSZoneFree([self zone], monitors);
+    }
   XCloseDisplay(dpy);
   [super dealloc];
 }
--- a/Source/x11/XGServer.m
+++ b/Source/x11/XGServer.m
@@ -141,7 +141,7 @@ _parse_display_name(NSString *name, int *dn, int *sn)
   XGDrawMechanism drawMechanism;
 }
 
-- initForDisplay: (Display *)dpy screen: (int)screen_number;
+- initForDisplay: (Display *)dpy screen: (int)screen_id;
 - (XGDrawMechanism) drawMechanism;
 - (RContext *) context;
 @end
@@ -173,7 +173,7 @@ _parse_display_name(NSString *name, int *dn, int *sn)
   return attribs;
 }
 
-- initForDisplay: (Display *)dpy screen: (int)screen_number
+- initForDisplay: (Display *)dpy screen: (int)screen_id
 {
   RContextAttributes *attribs;
   XColor testColor;
@@ -182,7 +182,7 @@ _parse_display_name(NSString *name, int *dn, int *sn)
    /* Get the visual information */
   attribs = NULL;
   //attribs = [self _getXDefaults];
-  rcontext = RCreateContext(dpy, screen_number, attribs);
+  rcontext = RCreateContext(dpy, screen_id, attribs);
 
   /*
    * If we have shared memory available, only use it when the XGPS-Shm
@@ -291,7 +291,7 @@ _parse_display_name(NSString *name, int *dn, int *sn)
       drawMechanism = XGDM_PORTABLE;
     }
   NSDebugLLog(@"XGTrace", @"Draw mech %d for screen %d", drawMechanism,
-        screen_number);
+        screen_id);
   return self;
 }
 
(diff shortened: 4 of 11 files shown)


Source/x11/XGServerWindow.m
```diff
@@ -4409,24 +4413,76 @@ - (void) freecursor: (void*) cid
...
+  
+  return screens;
+}
+#else // HAVE_XRANDR
+- (NSArray *)screenList
+{
+  int i;
+  NSMutableArray *screens;
+ 
+  monitorsCount = ScreenCount(dpy);
```
fredkiefer: Now monitorsCount is set, but monitors points to null, this is dangerous. Why not fill up the monitors array here too? The values for depth and resolution should work as above and for frame we could use 
boundsRect = NSMakeRect(0, 0, DisplayWidth(dpy, screen),
                           DisplayHeight(dpy, screen));
That way boundsForScreen: would be the same for both cases.

Source/x11/XGServerWindow.m
```diff
@@ -4409,24 +4413,76 @@ - (void) freecursor: (void*) cid
...
+    {
+      if (i != defScreen)
+        [screens addObject: [NSNumber numberWithInt: i]];
+    }
+  return screens;
 }
+#endif
 
-- (NSWindowDepth) windowDepthForScreen: (int) screen_num
+- (NSWindowDepth) windowDepthForScreen: (int) screen_num // X Screen - OK
```
fredkiefer: I really hate comments at the end of method definitions. If this is going to stay could you please move it before the method and make an explicit and understandable statement out of it?

fredkiefer:
What I am currently unsure is whether we should move more of the code over to XGServer.m the actual screenList ivar gets handled there and maybe it would be better to have everything that deals with XRandR and X screens in one place and leave just stuff that handles NSScreens over here? I really don't know. Most of that confusion is already in the old code, still we could try to clean it up a bit.

Source/x11/XGServerWindow.m
trunkmaster: Actually these are not comments but marks for me. I need to check end fix parameter name to be more clear what this method woud expect X11 screen number or NSScreen index. I'll remove it later.

Source/x11/XGServerWindow.m
trunkmaster: `monitors` used in RandR-related code only. Do you want to make use of monitors for both cases?

trunkmaster:
> What I am currently unsure is whether we should move more of the code over to XGServer.m the actual screenList ivar gets handled there and maybe it would be better to have everything that deals with XRandR and X screens in one place and leave just stuff that handles NSScreens over here? I really don't know. Most of that confusion is already in the old code, still we could try to clean it up a bit.

I'm thinking about move all of the screen related methods into seaparate file - XGServerScreen.m?
Plus rename all files into more convenient format <ClassName>+<CategoryName>.m? First time I looked at x11 code I think every file contain class implementation not category - it's confusing.
Also I'd like to make Xrandr library mandatory for backend and get rid of all these #ifdefs.

Source/x11/XGServerWindow.m
trunkmaster: I've finished fixing marked methods and removed marks.

Source/x11/XGServerWindow.m
trunkmaster: Sorry, I'd rather get rid of this code instead of enhance it. It looks overengineered: why we need to put defScreen at index 0? how many X screens it optimized for?

trunkmaster:
@fredkiefer after some extensive testing I've found that current implementation of screen handling is broken. I hope I'll find a solution during the weekend. I'll inform you after that.

Source/x11/XGServerWindow.m
```diff
@@ -4409,75 +4413,144 @@ - (void) freecursor: (void*) cid
   return depth;
 }
 
+/* This method assumes that we deal with one X11 screen - `defScreen`.
+   Basically it means that we have DISPLAY variable set to `:0.0`.
+   AppKit and X11 use the same term "screen" with different meaning:
+   AppKit Screen - single mintor/display device.
```
fredkiefer: monitor

Source/x11/XGServerWindow.m
```diff
@@ -4409,75 +4413,144 @@ - (void) freecursor: (void*) cid
...
+   X11 Screen - its a plane where X Server can draw on.
+   XRandR - Xorg extension that halps to manipulate monitors layout providing
```
fredkiefer: helps

Source/x11/XGServerWindow.m
```diff
@@ -4409,75 +4413,144 @@ - (void) freecursor: (void*) cid
   return depth;
 }
 
+/* This method assumes that we deal with one X11 screen - `defScreen`.
```
fredkiefer: Only in the XRandR case, for the fallback case multiple screens are possible.

Source/x11/XGServerWindow.m
```diff
@@ -4409,75 +4413,144 @@ - (void) freecursor: (void*) cid
...
+   Basically it means that we have DISPLAY variable set to `:0.0`.
+   AppKit and X11 use the same term "screen" with different meaning:
+   AppKit Screen - single mintor/display device.
+   X11 Screen - its a plane where X Server can draw on.
+   XRandR - Xorg extension that halps to manipulate monitors layout providing
+   X11 Screen. */
```
fredkiefer: I would say explicitly that we map XRandR monitors into NSScreens.

Source/x11/XGServerWindow.m
```diff
@@ -4409,75 +4413,144 @@ - (void) freecursor: (void*) cid
...
+
+      for (i = 0; i < monitorsCount; i++)
+        {
+          output_info = XRRGetOutputInfo(dpy, screen_res, screen_res->outputs[i]);
+          if (output_info->crtc)
+            {
+              crtc_info = XRRGetCrtcInfo(dpy, screen_res, output_info->crtc);
+
+              // Fill the cache of device parameters
+              monitors[i].depth = [self windowDepthForScreen: i];
```
fredkiefer: Why don't you pass defScreen here? That way the called methods could stay the same.
This will anyway result in all you monitors having the same depth and resolution, which seems wrong to me.

trunkmaster:
@fredkiefer, I'm finishing more comprehensive update to all this things. I hope I'll commit it today. Be There're huge number changes.

Source/x11/XGServerWindow.m
trunkmaster: Thanks, done.

Source/x11/XGServerWindow.m
trunkmaster: Thanks, done.

Source/x11/XGServerWindow.m
trunkmaster: Theoretically, yes. Not working in practice. I've wrote you in my mail.

Source/x11/XGServerWindow.m
trunkmaster: Thanks, done.

Source/x11/XGServerWindow.m
trunkmaster: Good, done.

trunkmaster:
@fredkiefer, sorry for such huge amount of changes. All of them was connected one by one...

Source/x11/XGServer.m
```diff
@@ -596,48 +615,48 @@ static int byte_order(void)
...
-      bpp = visualInfo->depth;
-      XFree(visualInfo);
-    }
-  else
-    {
-      visual = DefaultVisual(dpy, DefaultScreen(dpy));
-      bpp = DefaultDepth(dpy, DefaultScreen(dpy));
-    }
-#else
+//   /*
```
fredkiefer: Better remove this completely then to comment it out in yet another way. Now it would be commented out three times.

Source/x11/XGServer.m
```diff
@@ -515,56 +523,67 @@ - (Display *) xDisplay
...
+  return generic.appRootWindow;
+}
 
-  screen = NSMapGet(screenList, (void *)(uintptr_t)screen_number);
-  if (screen == NULL)
+- (XGScreenContext *) _defaultScreenContext
+{
+  XGScreenContext *screenCtxt;
+
+  screenCtxt = NSMapGet(screenList, (void *)(uintptr_t)defScreen);
```
fredkiefer: Why have a map if you only allow one screen?

Source/x11/XGServer.m
```diff
@@ -819,6 +821,9 @@ - (BOOL) setPreeditSpot: (NSPoint *)p
 #include <AppKit/NSClipView.h>
 #include <AppKit/NSTextView.h>
 
+#pragma clang diagnostic push
```
fredkiefer: This will generate useless warnings for gcc. Either make it conditional on clang or remove it again.

Source/x11/XGServerEvent.m
```diff
@@ -2683,16 +2682,16 @@ - (NSPoint) mouseLocationOnScreen: (int)screen window: (int *)win
...
+      if (screen >= 0 && screen != screen_id)
         {
           /* Mouse not on the requred screen, return an invalid point */
           return p;
         }
       height = attribs.height;
     }
   else
-    height = [self boundsForScreen: screen_number].size.height;
+    height = [self boundsForScreen: screen].size.height;
```
fredkiefer: In XGServerWindow.m you call this method with a monitor_id. Which one is correct?

Source/x11/XGServerWindow.m
```diff
@@ -1893,15 +1894,16 @@ - (int) window: (NSRect)frame : (NSBackingStoreType)type : (unsigned int)style
...
+  context = [self screenRContext];
 
   /* Create the window structure and set the style early so we can use it to
   convert frames. */
   window = NSAllocateCollectable(sizeof(gswindow_device_t), NSScannedOption);
   memset(window, '\0', sizeof(gswindow_device_t));
   window->display = dpy;
-  window->screen = screen;
+  window->screen_id = defScreen;
+  window->monitor_id = screen;
```
fredkiefer: Now you got me completely confused. Is the monitor_id that what in gui is the screen? Wouldn't it be clearer to name them the same, that is keep the screen name but refer to monitors in the XRandR case?

Source/x11/XGServerWindow.m
```diff
@@ -2096,15 +2098,16 @@ - (int) nativeWindow: (void *)winref : (NSRect*)frame : (NSBackingStoreType*)typ
   *screen = XScreenNumberOfScreen(win_attributes.screen);
```
fredkiefer: You don't seem to use this variable and this looks wrong to me. This is not a very common case, we rarely create a GNUstep window for an existing X window, but I am rather sure you broke that code.

Headers/x11/XGServerWindow.h
```diff
@@ -84,8 +84,9 @@ typedef struct _gswindow_device_t {
   Display               *display;      /* Display this window is on */
   Window                ident;         /* Window handle */
   Window                root;          /* Handle of root window */
-  Window		parent;        /* Handle of parent window */
-  int                   screen;        /* Screeen this window is on */
+  Window                parent;        /* Handle of parent window */
+  int                   screen_id;     /* Screeen this window is on */
```
fredkiefer: Why keep the screen_id? In your code this will only ever point to the default screen.

Source/x11/XGServerWindow.m
fredkiefer: I don't have a multi monitor setup, but others do. I would prefer to discuss this on the mailing list. I don't know whether it was possible to move a GNUstep window from one screen to another and how that was reflected in the windows data structure.

Source/x11/XGServerWindow.m
```diff
@@ -4409,75 +4415,140 @@ - (void) freecursor: (void*) cid
   return depth;
 }
 
+/* This method assumes that we deal with one X11 screen - `defScreen`.
+   Basically it means that we have DISPLAY variable set to `:0.0`.
```
fredkiefer: No, even if there could only be one screen, both numbers could be set to arbitrary values.

Source/x11/XGServerWindow.m
```diff
@@ -4409,75 +4415,140 @@ - (void) freecursor: (void*) cid
...
+   AppKit and X11 use the same term "screen" with different meaning:
+   AppKit Screen - single monitor/display device.
+   X11 Screen - it's a plane where X Server can draw on.
+   XRandR - Xorg extension that helps to manipulate monitor layout providing
+   X11 Screen. 
+   We map XRandR monitors (outputs) to NSScreen. */
 - (NSArray *)screenList
 {
-  /* I guess screen numbers are in order starting from zero, but we
```
fredkiefer: We still need to put the default monitor first, the gui code relies on that.

Source/x11/XGServerWindow.m
```diff
@@ -4409,75 +4415,140 @@ - (void) freecursor: (void*) cid
...
+            {
+              crtc_info = XRRGetCrtcInfo(dpy, screen_res, output_info->crtc);
+
+              // Fill the cache of device parameters
+              monitors[i].depth = [self windowDepthForScreen: defScreen];
+              monitors[i].resolution = [self resolutionForScreen: defScreen];
+              monitors[i].frame = NSMakeRect(crtc_info->x, crtc_info->y,
+                                             crtc_info->width, crtc_info->height);
+              // Add number of device
+              [tmpScreens addObject: [NSNumber numberWithInt: i]];
```
fredkiefer: That way we get a screen with the id 0, I am not sure whether our code will handle this correctly.

Source/x11/XGServerWindow.m
```diff
@@ -4409,75 +4415,140 @@ - (void) freecursor: (void*) cid
...
+      for (i = 0; i < monitorsCount; i++)
+        {
+          output_info = XRRGetOutputInfo(dpy, screen_res, screen_res->outputs[i]);
+          monitors[i].screen_id = defScreen;
+          if (output_info->crtc)
+            {
+              crtc_info = XRRGetCrtcInfo(dpy, screen_res, output_info->crtc);
+
+              // Fill the cache of device parameters
+              monitors[i].depth = [self windowDepthForScreen: defScreen];
```
fredkiefer: This looks wrong. If we have different monitors they all could have different depths and resolutions. Why set them all to the same value?

Source/x11/XGServerWindow.m
```diff
@@ -4409,75 +4415,140 @@ - (void) freecursor: (void*) cid
...
+                                     DisplayHeight(dpy, defScreen));
+      screens = [NSArray arrayWithObject: [NSNumber numberWithInt: defScreen]];
+    }
+  
+  return screens;
 }
 
-- (NSWindowDepth) windowDepthForScreen: (int) screen_num
+// `screen` is a Xlib screen number.
+- (NSWindowDepth) windowDepthForScreen: (int)screen
```
fredkiefer: This method will get called by gui as well and there this value will most likely be a monitor number. Please be aware that this method plus a few more are part of the public interface of back that gets used by MS Windows and the upcoming Wayland backend as well. We should not break these methods.

Source/x11/XGServerWindow.m
```diff
@@ -4409,75 +4415,140 @@ - (void) freecursor: (void*) cid
...
 
-  if (dpy == NULL)
+  if (dpy == NULL || screen < 0 || screen >= monitorsCount)
     {
       return NULL;
     }
 
-  screen = XScreenOfDisplay(dpy, screen_num);
-  if (screen == NULL)
+  x_screen = XScreenOfDisplay(dpy, monitors[screen].screen_id);
```
fredkiefer: How could that in your code ever be different from the default screen?

Headers/x11/XGServer.h
```diff
@@ -62,21 +69,23 @@ typedef enum {
   id                inputServer;
   int               randrEventBase;
   int               randrErrorBase;
+  MonitorDevice     *monitors;
+  unsigned          monitorsCount;
 }
 
-+ (Display*) currentXDisplay;
-- (Display*) xDisplay;
++ (Display *) xDisplay;
```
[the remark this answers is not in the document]
trunkmaster: I think it's not confusing. `+currentXDisplay` calls `-xDisplay`. So basically it's the same `dpy`. No matter what method (class or instace) you're calling - you will get the same `Display *`.

Source/x11/XGServerWindow.m
trunkmaster: I've made some additional testing (you can do it too using VirtualBox with "Monitor Count" set to 2 in VM preferences). I've made testing with `DISPLAY=:0.0`, `DISPLAY=:0`, with enable and disabled XRandR extension (`Option "RANR" "Disabled"` in xorg.conf). In all cases `ScreenCount` returns `1`.

Let me explain once more: GNUstep application can get only 1 (one) screen at a time. Screen identifier (XID) is determined at X11 initialization in backend by parsing DISPLAY environment variable and stored in `defScreen` (see _initXContext in XGServer.m). With current backend implementation (without XRandR support) there's no any chance to get multiple screens (Xlib screens).

What question/topic do we need to discuss in mailing list? I'm confused.

Headers/x11/XGServer.h
```diff
@@ -62,21 +69,23 @@ typedef enum {
...
+- (Display *) xDisplay;
+- (Window) xDisplayRootWindow;
 - (Window) xAppRootWindow;
 
-- (void *) xrContextForScreen: (int)screen_number;
-- (Visual *) visualForScreen: (int)screen_number;
-- (int) depthForScreen: (int)screen_number;
+- (void *) screenRContext;
```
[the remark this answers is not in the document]
trunkmaster: If somebody wants multiple monitors support in modern Xorg - XRandR extension is mandatory. You may disagree with me but reality is like that. Look at KDE, GNOME, Xfce. We need to get rid of obsolete (and broken) code and design approach if GNUstep pretends to be modern framework.

Source/x11/XGServer.m
trunkmaster: You're right, done.

Source/x11/XGServer.m
trunkmaster: I'd prefer not to touch this until we agreed on "single Xlib screen" concept.

Source/x11/XGServer.m
trunkmaster: Good point. Fixed.

Source/x11/XGServerEvent.m
trunkmaster: You're right, it should be monitor_id. I've not figured out how we can get monitor_id here. It's in my todo list.

Source/x11/XGServerWindow.m
trunkmaster: > Now you got me completely confused. Is the monitor_id that what in gui is the screen? 

Yes.

>Wouldn't it be clearer to name them the same, that is keep the screen name but refer to monitors in the XRandR case?

screen_id used in multiple places as Xlib screen for Xlib function calls - we should keep that structure element.

Source/x11/XGServerWindow.m
trunkmaster: I use it in Workspace's RecyclerIcon through NSWindow's initWithWindowRef: - works correctly. But you're right I need to assign this value to screen_id instead of monitor_id.

Headers/x11/XGServerWindow.h
trunkmaster: screen_id used in multiple places as Xlib screen for Xlib function calls - we should keep that structure element.

Source/x11/XGServerWindow.m
trunkmaster: You're right it's not clear description. I mean for single application copy we have signle display and scree. I'll change desciption.

Source/x11/XGServerWindow.m
trunkmaster: We have single Xlib screen no matter how much monitors we have. That's why.

Source/x11/XGServerWindow.m
trunkmaster: You're right. I'll fix this.

Source/x11/XGServerWindow.m
trunkmaster: Fixed.

Source/x11/XGServerWindow.m
trunkmaster: Done.

Source/x11/XGServerWindow.m
trunkmaster: You're right. Let's leave this as is for now. I'll remove it later if we decide to remove gswindow_device_t->screen_id.

Source/x11/XGServerWindow.m
trunkmaster: Done.

Source/x11/XGServerWindow.m
trunkmaster: I see no problem here. Our code handles it correctly.

Source/x11/XGServerWindow.m
trunkmaster: Done.

Source/x11/XGServerWindow.m
```diff
@@ -4409,75 +4415,152 @@ - (void) freecursor: (void*) cid
...
+        }
+      XRRFreeScreenResources(screen_res);
+      screens = [NSArray arrayWithArray: tmpScreens];
+    }
+#endif
+
+  if (monitorsCount == 0)
+    {
+      /* Assumed that it's always available 1 screen per application. We only
+         need to know it's number and it was saved in _initXContext as
```
fredkiefer: The second "it's" should be "its". This gets simpler if you turn the first one into "it is", or rather "there is" :-)

Source/x11/XGServerWindow.m
```diff
@@ -4409,75 +4415,152 @@ - (void) freecursor: (void*) cid
...
+      int depth = x_screen->depths[index].depth;
       depths[index] = _computeDepth(class, depth);
     }
   depths[index] = 0; // terminate with a zero.
 
   return depths;
 }
 
-- (NSSize) resolutionForScreen: (int)screen_num
+// `screen` is a Xlib screen number.
```
fredkiefer: Wouldn't the name screen_num be better in that case?

trunkmaster: Agree. Thanks, fixed.

Source/x11/XGServerWindow.m
trunkmaster: Good point, thanks. Fixed.

Source/x11/XGServerEvent.m
trunkmaster: I've checked this code - it's correct. mouseLocationOnScreen:window: correctly called by GUI classes with `screen` parameter set to NSScreen's screenNumber.

fredkiefer: And NSScreens screenNumbers now are monitor_ids.

trunkmaster: Exactly. And boundsForScreen: waits for monitor_id.

ChangeLog
```diff
@@ -1,21 +1,134 @@
-2020-92-07 Fred Kiefer <FredKiefer@gmx.de>
+2020-02-07 Fred Kiefer <FredKiefer@gmx.de>
```
fredkiefer: Thank you for fixing this!

ChangeLog
```diff
@@ -1,21 +1,134 @@
...
+	XGServer's method.
+
+
 2020-01-31  Sergii Stoian  <stoyan255@gmail.com>
 
 	* Source/x11/XGServerWindow.m (boundsForScreen:): check for number of
 	XRandR's ouputs.
 
-2010-01-31  Riccardo Mottola <rm@gnu.org>
+2020-01-31  Riccardo Mottola <rm@gnu.org>
```
fredkiefer: And this :-)

Source/x11/XGServerEvent.m
```diff
@@ -1045,7 +1046,12 @@ - (void) processEvent: (XEvent *) event
 		 * the programa can move/resize the window while we send
 		 * this event, causing a confusion.
 		 */
-		[[NSApp windowWithWindowNumber: cWin->number] sendEvent: r];
+                window = [NSApp windowWithWindowNumber: cWin->number];
```
fredkiefer: Indentation of this line looks wrong.

Source/x11/XGServerEvent.m
```diff
@@ -1045,7 +1046,12 @@ - (void) processEvent: (XEvent *) event
...
+		[window sendEvent: r];
+                /* Update monitor_id of the backend window.
+                   NSWindow may change screen pointer while processing
+                   the event. */
+                cWin->monitor_id = [[window screen] screenNumber];
```
fredkiefer: Computing the screen of a window may be a rather expensive operation. Do we really want to execute this after every event?

Source/x11/XGServerWindow.m
```diff
@@ -576,7 +576,7 @@ - (NSRect) _XFrameToOSFrame: (NSRect)x for: (void*)window
 
   [self styleoffsets: &l : &r : &t : &b : style : win->ident];
   o = x;
-  o.origin.y = [self boundsForScreen: win->screen].size.height - x.origin.y;
+  o.origin.y = [self boundsForScreen: win->monitor_id].size.height - x.origin.y;
```
fredkiefer: Could you explain why you use the height of the monitor here while you used the screen height in _OSFrameToXFrame: These two methods should be symmetrical.

Source/x11/XGServerWindow.m
```diff
@@ -3190,7 +3192,7 @@ - (void) movewindow: (NSPoint)loc : (int)win
     }
 
   window->siz_hints.x = (int)loc.x;
-  window->siz_hints.y = (int)([self boundsForScreen: window->screen].size.height -
+  window->siz_hints.y = (int)([self boundsForScreen: window->monitor_id].size.height -
```
fredkiefer: The implementation here and in the next method needs to match _XFrameToOSFrame:, currently it does but it may be that all these are wrong. If the screen is bigger than the current monitor, this results in wrong window placements.

Source/x11/XGServerWindow.m
fredkiefer: I think that we left enough time for anybody to reply on the mailing list. As there was no response I withdraw my comment.

Source/x11/XGServerWindow.m
```diff
@@ -4430,74 +4436,154 @@ - (void) freecursor: (void*) cid
...
+      for (i = 0, mi = 0; i < screen_res->noutput; i++)
+        {
+          output_info = XRRGetOutputInfo(dpy, screen_res, screen_res->outputs[i]);
+          if (output_info->crtc)
+            {
+              crtc_info = XRRGetCrtcInfo(dpy, screen_res, output_info->crtc);
+              
+              monitors[mi].screen_id = defScreen;
+              monitors[mi].depth = [self windowDepthForScreen: mi];
+              monitors[mi].resolution = [self resolutionForScreen: defScreen];
```
fredkiefer: You could compute this once before the loop. And it still looks wrong to me. Why would different monitors have the same resolution?

Source/x11/XGServerWindow.m
```diff
@@ -4430,74 +4436,154 @@ - (void) freecursor: (void*) cid
...
+
+  if (monitorsCount == 0)
+    {
+      /* It is assumed that there is always only one screen per application. 
+         We only need to know its number and it was saved in _initXContext as
+         `defScreen`. */
+      monitorsCount = 1;
+      monitors = NSZoneMalloc([self zone], sizeof(MonitorDevice));
+      monitors[0].screen_id = defScreen;
+      monitors[0].depth = [self windowDepthForScreen: defScreen];
```
fredkiefer: Does this method get called with a monitor or a screen id?

Source/x11/XGServerWindow.m
fredkiefer: Now it is even worse. This method gets called when setting up the monitor list. In the XRandR case it gets a screen id, in the old case a monitor id and the list of the monitors is not set up. In your view of the world this method should ignore its parameter and always return the depth of the default screen. And this is in my view wrong as not all monitors have the same depth.

fredkiefer:
I went through the whole code once again and there are still a few places where I see a bit of confusion. With that resolved I would be fine with merging this.

Source/x11/XGServerWindow.m
trunkmaster: The code we're commenting seems outdated. Look here https://github.com/gnustep/libs-back/blob/bcf71f230214873a9621abf295d344a3c8603f69/Source/x11/XGServerWindow.m#L4527. At least my comment in code differs. It states "`screen` is a monitor index not X11 screen". This method should always receive monitor index. As for non-XRandR case monitor list gets set up now (look at bottom of -screenList method). Although I'll fix -windowDepthForScreen call for that case (should send `0` as parameter).

ChangeLog
trunkmaster: :-)

ChangeLog
trunkmaster: :-)

Source/x11/XGServerEvent.m
trunkmaster: Tabs... Fixed.

Source/x11/XGServerWindow.m
trunkmaster: Thanks, fixed.

Source/x11/XGServerWindow.m
trunkmaster: Thanks, fixed.

Source/x11/XGServerWindow.m
trunkmaster: Indeed it's wrong. But resolutionForScreen is not implemented yet correctly. It's always returns {72, 72}. I've already said - it's rather complicated separate task and involves -back and -gui changes. I'll change this after correct resolutionForScreen: implemenation.

Source/x11/XGServerWindow.m
trunkmaster: Monitor. In this case it should be `0`. Fixed.

Source/x11/XGServerEvent.m
trunkmaster: Hm... I need to think about it.

trunkmaster:
@fredkiefer, I plan to finish my "app menu follows key window" implementation (mostly in -gui). It's almost done, but has some flaws. Probably more fixes will go into -back.

Source/x11/XGServerWindow.m
fredkiefer: Yes, you are right, for some reasons I did see old code here

Source/x11/XGServerWindow.m
fredkiefer: Now the commented out code uses the wring variable name. But no harm in that.

trunkmaster: For the sake of perfection I've fixed it. :-)

trunkmaster:
Just for the record, useful changes would be:
- do not change monitor_id on every window move;
- do not use NSMapTable for Xlib screen if we will decide to support only XRandR capable environments;
- implement resolutionForScreen: method.

PR #20, opened 2020-01-30
Fix: right-click app menu steals focus from active application

trunkmaster:
This change fixes focus stealing from active application on right-click menu appearance.
How to reproduce:
1. Start two applications (app1 and app2).
2. Activate app1 by double-click on its appicon.
3. Right-click on app2 appicon without activating it.

App2 becomes active, right click main menu shows. After right mouse button release app2 stays active.

Should be: app2 right-click application menu shows without app2 activation. After right mouse release app1 stays active.

### DIFF
--- a/Source/x11/XGServerEvent.m
+++ b/Source/x11/XGServerEvent.m
@@ -1981,6 +1981,13 @@ posixFileDescriptor: (NSPosixFileDescriptor*)fileDescriptor
       NSDebugLLog(@"Focus", @"Ignoring window focus request");
       cWin->ignore_take_focus = NO;
     }
+  else if ([[NSApp mainMenu] isTransient] != NO)
+    {
+      /* Do not grab focus from active application if right-click on our
+         application icon was performed. */
+      NSDebugLLog(@"Focus",
+                  @"Ignore transient application menu focus request.");
+    }
   else if (cWin->number == key_num)
     {
       NSDebugLLog(@"Focus", @"Reasserting key window");
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2020-01-30  Sergii Stoian  <stoyan255@gmail.com>
+
+	* Source/x11/XGServerEvent.m (_handleTakeFocusAtom:forContext:):
+	do not grab focus of active applicaion if right-click app menu requests
+	focus.
+
 2020-01-26  Sergii Stoian  <stoyan255@gmail.com>
 
 	* Source/x11/XGServerWindow.m (boundsForScreen:): use `screen` variable


Source/x11/XGServerEvent.m
```diff
@@ -1981,6 +1981,13 @@ - (NSEvent *)_handleTakeFocusAtom: (XEvent)xEvent
       NSDebugLLog(@"Focus", @"Ignoring window focus request");
       cWin->ignore_take_focus = NO;
     }
+  else if ([[NSApp mainMenu] isTransient] != NO)
```
fredkiefer: Shouldn't this only be checked if cWin->number == [[[NSApp mainMenu] window] windowNumber]? We could have clicked on a completely different window.

trunkmaster: No, actually. No matter what window of inactive application we have made right-click. Application menu shouldn't be visible and as a result shouldn't receive TakeFocus request.

fredkiefer: I just don't understand this. Your comment in the code states that a right click on the application icon happened. How do you know that this was a right click and that the application is inactive?
The only thing we know here is that the application is displaying its main menu transiently. We do not know where the click happened. The click could even be on the menu itself or on a submenu. (You handle that case from line 1997 on) Or it may be any other window of the application with a text field in it.

trunkmaster: > I just don't understand this. Your comment in the code states that a right click on the application icon happened. How do you know that this was a right click and that the application is inactive?

I understand your confusion - it's my fault. Comment is not accurate enough. I've wrote that comment when observed wrong behavior after right-click on appicon. After patch creation I realize that it is the general case when transient main menu appeared.

> The only thing we know here is that the application is displaying its main menu transiently. We do not know where the click happened. The click could even be on the menu itself or on a submenu. (You handle that case from line 1997 on) Or it may be any other window of the application with a text field in it.

Correct. Transient main menu opened on on right-click no matter what window was clicked. That is the case.

trunkmaster:
@fredkiefer ?

Source/x11/XGServerEvent.m
fredkiefer: OK, let me try a different approach to explain my issue with the code. Forget what you are intending to do. Just look at the code and see what it actually does.
When ever there is a transient main menu open all further mouse clicks won't transfer the focus to a new window. That is you are working in one window typing in stuff, you decide to open the main window by right clicking but then decide that you rather want to work on another window and (left) click on that. This click will no longer transfer the focus to the new window. Is this your intention?

trunkmaster: > OK, let me try a different approach to explain my issue with the code. Forget what you are intending to do. Just look at the code and see what it actually does.
> When ever there is a transient main menu open all further mouse clicks won't transfer the focus to a new window. That is you are working in one window typing in stuff, you decide to open the main window by right clicking but then decide that you rather want to work on another window and 

... you release right mouse button and...

> (left) click on that. This click will no longer transfer the focus to the new window. 

Wrong. Do you know how right-click menu in GNUstep work?

> Is this your intention?

Sure not!

trunkmaster:
It seems that review process got stuck and this issue is only mine. Move it to [NEXTSPACE](https://github.com/trunkmaster/nextspace/tree/master/Libraries/gnustep/back-art). Closing this.

PR #21, opened 2020-02-11
Align screens origin to OpenStep coordinate system.

trunkmaster:
Transform screen origin into unflipped (OpenStep) coordinate system for XRandR capable systems. After this change windows and screen will use the same start of coordinate system for their origins.

### DIFF
--- a/Source/x11/XGServerWindow.m
+++ b/Source/x11/XGServerWindow.m
@@ -4465,6 +4465,7 @@ _computeDepth(int class, int bpp)
       NSMutableArray *tmpScreens;
       int             i;
       int             mi;
+      int             xScreenHeight = DisplayHeight(dpy, defScreen);
 
       monitorsCount = screen_res->noutput;
       tmpScreens = [NSMutableArray arrayWithCapacity: monitorsCount];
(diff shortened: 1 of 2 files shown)


Source/x11/XGServerWindow.m
```diff
@@ -4483,8 +4485,13 @@ - (NSArray *)screenList
               monitors[mi].screen_id = defScreen;
               monitors[mi].depth = [self windowDepthForScreen: mi];
               monitors[mi].resolution = [self resolutionForScreen: defScreen];
-              monitors[mi].frame = NSMakeRect(crtc_info->x, crtc_info->y,
-                                              crtc_info->width, crtc_info->height);
+              /* Transform coordinates from Xlib (flipped) to OpenStep (unflippped). 
+                 Windows and screens should have the same coordinate system. */
+              frame = NSMakeRect(crtc_info->x,
```
fredkiefer: You don't need this intermediate variable.

trunkmaster:
> You should merge in master first. I already did some clean up on that branch that needs to be included here.

Great, thank you. But you've broke the logic of code. The old code logic was: if XRandR failed to get screen info (screen_re->noutput == 0) we failback to old approach - one screen and its Xlib screen. But I'll fix it later.

Source/x11/XGServerWindow.m
trunkmaster: I use it to make code width less than 80 characters. Anyway I've made changes.

PR #22, opened 2020-02-23
Window positionning: rely on internal variable not the GUI's one

trunkmaster:
Using `window->xframe` for making decisions provides ability to alter NSWindow's _frame variable before `placewindow:for:` call without affecting actual Xlib window placement.


Source/x11/XGServerWindow.m
```diff
@@ -3235,7 +3235,7 @@ - (void) placewindow: (NSRect)rect : (int)win
   NSDebugLLog(@"XGTrace", @"DPSplacewindow: %@ : %d", NSStringFromRect(rect),
 	      win);
   nswin  = GSWindowWithNumber(win);
```
fredkiefer: Maybe we should move this line further down as nswin is no longer needed here and in many case it will never be needed at all.

trunkmaster: Sure, done.

PR #24, opened 2020-03-05
placewindow:: fix frame check for multi-monitor layout 

trunkmaster:
When monitors layout was changed windows need to update their positions on screen. There are the cases when OpenStep frames is not changed while Xlib did and vise versa. This fix adds checks for OpenStep frame changes as well as Xlib's.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2020-03-05  Sergii Stoian  <stoyan255@gmail.com>
+
+	* Source/x11/XGServerWindow.m (placewindow::): check window frame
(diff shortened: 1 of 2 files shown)


trunkmaster:
> Sorry, I lost track of these different changes. This one seems to revert some previous change plus to add on top. Hopefully you understand what you are doing :-)

Sure I know. The changes are quite simple:
1. Use internal cached `frame` of NSWindow (OpenStep) instead of NSWindow's frame in if statements.
2. Also use Xlib cached frame of window in if statements.

PR #25, opened 2020-03-11
XGServerWindow.m small fixes

trunkmaster:
I've checked, 24-bit depth works good for images with alpha channel (32-bit depth).
Second fix reduce number of cases when "styleoffsets ... guessing offsets" messages appears.

### DIFF
--- a/Source/x11/XGServerWindow.m
+++ b/Source/x11/XGServerWindow.m
@@ -2283,7 +2283,8 @@ _get_next_prop_new_event(Display *display, XEvent *event, char *arg)
 
   if ((style & NSIconWindowMask) || (style & NSMiniWindowMask))
     {
-      style = NSBorderlessWindowMask;
+      *l = *r = *t = *b = 0.0;
+      return;
     }
 
   /* Next try to get the offset information that we have obtained from
@@ -2841,7 +2842,7 @@ swapColors(unsigned char *image_data, NSBitmapImageRep *rep)
   height = [rep pixelsHigh];
   colors = [rep samplesPerPixel];
 
-  if (rcontext->depth != 32)
+  if (rcontext->depth != 32 && rcontext->depth != 24)
     {
       NSLog(@"Unsupported context depth %d", rcontext->depth);
       return 0;
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2020-03-11  Sergii Stoian  <stoyan255@gmail.com>
+
+	* Source/x11/XGServerWindow.m (_createAppIconPixmaps): accept
+	WRaster context depth 24.
+	(styleoffsets::::::): do not guess offsets for miniwindow and
+	appicon they are equal to 0.0.
+
 2020-03-05  Sergii Stoian  <stoyan255@gmail.com>
 
 	* Source/x11/XGServerWindow.m (placewindow::): check window frame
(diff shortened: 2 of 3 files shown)


Source/x11/XGServerWindow.m
```diff
@@ -2283,7 +2283,8 @@ - (void) styleoffsets: (float *) l : (float *) r : (float *) t : (float *) b
 
   if ((style & NSIconWindowMask) || (style & NSMiniWindowMask))
     {
-      style = NSBorderlessWindowMask;
+      *l = *r = *t = *b = 0.0;
```
fredkiefer: Why do you think this required? Shouldn't we first check what is already stored in the known offsets?
There may be window managers that put a border even on these windows.

Source/x11/XGServerWindow.m
```diff
@@ -2841,7 +2842,7 @@ - (int) _createAppIconPixmaps
   height = [rep pixelsHigh];
   colors = [rep samplesPerPixel];
 
-  if (rcontext->depth != 32)
+  if (rcontext->depth != 32 && rcontext->depth != 24)
```
fredkiefer: I did put this new test here on purpose as your code did not work for Riccardo. Please have him test this change first. We do not need a regression shortly before a release.

Source/x11/XGServerWindow.m
trunkmaster: There are 2 things here:
1. It's a special windows: miniwindow and appicon. If window manager defines offsets for such type of windows user should better disable appearance of them via defaults.
2. Simple application has 2 windows which produce "styleoffsets ... guessing offsets" messages: appicon and window with ID == 0. I see exacly 2 messages at application start. It means that code at line range 2295-2306 doesn't work either. Finally in WindowMaker we will end up with offsets 1.0 and that's wrong for miniwindow and appicon. For other WMs it will be 4.0. Is it what we need for our appicon and miniwindows?

In general I think that code below NSLog line is wrong and useless. Why offset for EWMH-compliant WM is 4.0? Why titlebar and resizebar sizes fixed and defined to that values?

Source/x11/XGServerWindow.m
trunkmaster: OK, no problem.

rmottola: I tested this specific change on both the problematic T23 16bit as well as the Letux400 (16 bit too) with no adverse effects.
However, I noticed one big thing: possibly due to this skipping, on the Letux app mini windows are not displayed anymore. Like battery monitor is unable to display the battery, only the icon is displayed. I fear this is because all the code is skipped because now it works only in certain conditions, while before it worked.

fredkiefer: Just to clarify, the change that Sergii made tries to allow for more systems to use this new code. It was my change to check for the depth that shut of Letux. I did this because you got a segmentation fault with the new code. If you don't get a segmentation fault now things are improved and Sergii's change to allow for depth 24 may be correct. But it is more likely that one of the other checks I added still blocks the call. You should check for that first.

fredkiefer:
It is probably best to split this up into three separate PRs and discuss each separately.

fredkiefer:
I think we should split up this PR into smaller parts. Some of this is already addressed by changes done by Thalion in other PRs and overall it was never a great idea to have unrelated changes in one PR.

PR #28, opened 2020-09-16
XGServerEvent.m: unhide app if WindowMaker sends WM_TAKE_FOCUS to window

trunkmaster:
If application (window) receives WM_TAKE_FOCUS in hidden state it means WindowMaker wants us to be unhidden.

Description: There are 2 protocols defined in XGServerWindow.h - WMFHideApplication and WMFHideOtherApplications. These protocols exist to help GNUstep application and WindowMaker notify each other about "Hide" and "Hide Others" actions. There are no protocols for "unhide" action.

Test case steps:
  1. Open application with window opened.
  2. Hide application with "Hide" menu item.
  3. Open WindowMaker's "Windows" menu (middle-click on desktop).
  4. Select hidden application window.
 Application unhides, activates, restored window is focused (receives) input and "Hide" menu item unselected.

### DIFF
--- a/Source/x11/XGServerEvent.m
+++ b/Source/x11/XGServerEvent.m
@@ -1983,7 +1983,7 @@ posixFileDescriptor: (NSPosixFileDescriptor*)fileDescriptor
              state - it's time to unhide. There's no other method to
              tell us to unhide. */
           NSDebugLLog(@"Focus", @"WM take focus while hidden - unhiding.");
-          [NSApp unhide:nil];
+          [NSApp unhide: nil];
         }
       else
         {
--- a/Source/x11/XGServerEvent.m
+++ b/Source/x11/XGServerEvent.m
@@ -1977,10 +1977,21 @@ posixFileDescriptor: (NSPosixFileDescriptor*)fileDescriptor
      events */
   if ([NSApp isHidden])
     {
-      /* This often occurs when hidding an app, since a bunch of
-         windows get hidden at once, and the WM is searching for a
-         window to take focus after each one gets hidden. */
-      NSDebugLLog(@"Focus", @"WM take focus while hiding");
+      if (generic.wm & XGWM_WINDOWMAKER)
+        {
+          /* If window receives WM_TAKE_FOCUS and application is in hidden
+             state - it's time to unhide. There's no other method to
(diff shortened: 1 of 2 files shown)


Source/x11/XGServerEvent.m
```diff
@@ -1977,10 +1977,21 @@ - (NSEvent *)_handleTakeFocusAtom: (XEvent)xEvent
...
-         windows get hidden at once, and the WM is searching for a
-         window to take focus after each one gets hidden. */
-      NSDebugLLog(@"Focus", @"WM take focus while hiding");
+      if (generic.wm & XGWM_WINDOWMAKER)
+        {
+          /* If window receives WM_TAKE_FOCUS and application is in hidden
+             state - it's time to unhide. There's no other method to
+             tell us to unhide. */
+          NSDebugLLog(@"Focus", @"WM take focus while hidden - unhiding.");
+          [NSApp unhide:nil];
```
fredkiefer, line 1986: Could you please add a space before "nil"?

fredkiefer (approved):
I understand what you want to achieve with the patch, but when I saw your original change I was worried about possible side effects as the old code has a quite clear comment on why it is there.
With your restriction on WindowMake this patch should be less dangerous. I sill would suggest that you test it with an application that has multiple windows and see if that hides correctly.

Source/x11/XGServerEvent.m
trunkmaster: Done.

trunkmaster:
Sure I've made numerous tets. Application hiding performed by WindowMaker. The actual call sequence is the following:
- `[NSWindow sendEvent]` - lets assume it was key press (Cmd-h)
- `[XGServer hideApplication]` - for WindowMaker it returns `YES` so GNUstep doesn't perform any hiding actions
- `[XGServer _sendRoot:typewindow:data0:data1:data2:data3:]` - sends WMFHideApplication to WindowMaker
After that WindowMaker unmaps all application's windows and set focus to next in window focus list ([wHideApplication()](https://github.com/window-maker/wmaker/blob/master/src/actions.c#L1714)).

As you can see WindowMaker definitely is not the case when "WM is searching for a window to take focus after each one gets hidden".

PR #29, opened 2020-09-18
x11/XGServerWindow.m: removed WindowMaker appicon hack

trunkmaster:
The removed code removed creates and maps zero-sized window in hope that WindowMaker will recognize app by this event and  create app icon window. Although when application requests window for appicon through the `-window::::` method call, it returns newly created and mapped window. This actually happens but application icon flickers between WM icon creation and application icon window appearance.

This fix based on the fact that root application window (ROOT) is mapped in `-orderwindow:::` and this is the event of application recognition by WindowMaker because it is a first application window that is mapped.

I've made tests on WindowMaker and NextSpace Workspace WM with ART and Cairo backends. Applications started with double-click is started with 'highlighted' icon and then normal icon without interim state - empty tile background with little white square at the center. It looks as more smooth application startup.

### DIFF
--- a/Source/x11/XGServerWindow.m
+++ b/Source/x11/XGServerWindow.m
@@ -77,8 +77,6 @@
 
 
 static BOOL handlesWindowDecorations = YES;
-// static int _wmAppIcon = -1;
-
 
 #define WINDOW_WITH_TAG(windowNumber) (gswindow_device_t *)NSMapGet(windowtags, (void *)(uintptr_t)windowNumber)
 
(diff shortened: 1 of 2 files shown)


Source/x11/XGServerWindow.m
```diff
@@ -77,7 +77,7 @@
 
 
 static BOOL handlesWindowDecorations = YES;
-static int _wmAppIcon = -1;
+// static int _wmAppIcon = -1;
```
fredkiefer, line 80: Please remove the code completely before merging this patch.

Source/x11/XGServerWindow.m
```diff
@@ -1606,16 +1606,16 @@ - (void) _setupRootWindow
    * generates some temporary windows to determine the window frame offsets
    * and reuse the icon window once the real app icon window is allocated.
    */
```
fredkiefer, line 1608: Please remove this comment as well.

Source/x11/XGServerWindow.m
```diff
@@ -1884,16 +1884,16 @@ - (int) window: (NSRect)frame : (NSBackingStoreType)type : (unsigned int)style
   /* WindowMaker hack: Reuse the empty app icon allocated in _setupRootWindow
    * for the real app icon.
    */
```
fredkiefer, line 1886: Again the code and the comment need to be removed completely.

fredkiefer (approved):
As this change affects only WindowMake behaviour I completely trust you that this improve the user experience.

Please don't just comment out the code but remove it completely instead.

trunkmaster:
Of course, it's my fault! Commented out code was removed.

PR #31, opened 2021-03-03
Win64 fixes

rmottola:
Seems all right to me !

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,20 @@
+2021-03-03 Riccardo Mottola <rm@gnu.org>
+
+	* Source/win32/WIN32Server.m:
+	actually store in a LONG_PTR the result of GetWindowLongPtr
+
+2021-03-02 Riccardo Mottola <rm@gnu.org>
+
+	* Source/win32/WIN32Server.m
+	* Source/win32/w32_create.m
+	* Source/win32/w32_general.m
+	* Source/win32/w32_movesize.m
+	* Source/win32/w32_windowdisplay.m
+	* Source/winlib/WIN32GState.m
+	* Source/cairo/Win32CairoGlitzSurface.m:
+	Consistently use 64bit safe SetWindowLongPtr/GetWindowLongPtr instead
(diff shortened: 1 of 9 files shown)


ChangeLog
```diff
@@ -1,3 +1,20 @@
+2021-03-03 Riccardo Mottola <rm@gnu.org>
```
fredkiefer, line 1: When doing a pull request it is better to squash the ChangeLog entries into one.

Source/win32/WIN32Server.m
```diff
@@ -1635,19 +1635,19 @@ - (int) window: (NSRect)frame : (NSBackingStoreType)type : (unsigned int)style
...
-        LONG    wstyleOld  = GetWindowLong(hwnd, GWL_STYLE);
-        LONG    estyleOld  = GetWindowLong(hwnd, GWL_EXSTYLE);
-        LONG    wstyleNew  = (wstyleOld & ~WS_OVERLAPPEDWINDOW);
-        LONG    estyleNew  = estyleOld | WS_EX_TOOLWINDOW;
+        LONG_PTR  wstyleOld  = GetWindowLongPtr(hwnd, GWL_STYLE);
+        LONG_PTR  estyleOld  = GetWindowLongPtr(hwnd, GWL_EXSTYLE);
+        LONG_PTR  wstyleNew  = (wstyleOld & ~WS_OVERLAPPEDWINDOW);
+        LONG_PTR  estyleNew  = estyleOld | WS_EX_TOOLWINDOW;
         
         NSDebugMLLog(@"WCTrace", @"wstyles - old: %8.8X new: %8.8X\n",
```
fredkiefer, line 1643: Probably we now need to use "lX" here instead of "X".

fredkiefer (approved):
Having participated in the original change, I am surely not able to review this unbiased. Still I am sure this change is a huge improvement over our previous code for WIn64. There will be plenty of places where more fixes are needed and I hope somebody will address these over the next months.

PR #33, opened 2021-11-06
Wayland backend xdg-surface and wrl-layer

nongio:


Source/cairo/WaylandCairoSurface.m
```diff
@@ -176,62 +184,41 @@ - (void) dealloc
...
-    NSDebugLog(@"updating region: %dx%d %dx%d", x, y, width, height);
-
-    if (cairo_surface_status(cairo_surface) != CAIRO_STATUS_SUCCESS)
-    {
-	NSWarnMLog(@"cairo initial window error status: %s\n",
-		   cairo_status_to_string(cairo_surface_status(_surface)));
+    window->buffer_needs_attach = YES;
+    if (window->configured) {
+        window->buffer_needs_attach = NO;
+        wl_surface_attach(window->surface, window->buffer, 0, 0);
```
fredkiefer, line 211: Do we need the attach call here even when it was already called in -initWithDevice:?

I would expect that we have a flag that tells us, whether this was already called if there is no consistent order in which these methods get called.

Source/wayland/WaylandServer+Cursor.m
```diff
@@ -0,0 +1,564 @@
...
+    /* FIXME: unlike in _motion and _axis handlers, the argument used in _button
+       is the "serial" of the event, not passed and unavailable in _motion and
+       _axis handlers. Is it allowed to pass "serial" as the eventNumber: in
+       _button handler, but "time" as the eventNumber: in the _motion and _axis
+       handlers? */
+    tick = serial;
+
+    NSDebugLog(@"sending pointer event at: %fx%f, window=%d", wlconfig->pointer.x, wlconfig->pointer.y, window->window_id);
+
+    /* FIXME: X11 backend uses the XGetPointerMapping()-returned values from
```
fredkiefer, line 311: We also support the version of the event without the button number. Just use that if you prefer. The more important bit is to set the event type correctly as for the first three buttons we already have the information there.

fredkiefer:
This is way too much code for me to read through. As Wayland support is still experimental in GNUstep I would say we just take your code as it seems to work better than the previous one.
It would be great if you could clean up the code a bit. That is remove commented out portions and try to stick to the GNUstep coding standard. (See http://www.gnustep.org/resources/documentation/Developer/CodingStandards/coding-standards.ps) Yes, this is horrible and really old. But having some standard saves us from discussing this every year.

nongio:
thanks for checking it @fredkiefer I updated the code using clang-format with this configuration file:
https://github.com/gnustep/libs-base/blob/master/.clang-format
let me know if it can go like this

Headers/wayland/WaylandServer.h
```diff
@@ -5,7 +5,8 @@
    Copyright (C) 2020 Free Software Foundation, Inc.
 
    Author: Sergio L. Pascual <slp@sinrega.org>
-   Date: February 2016
```
fredkiefer, line 8: You should leave the old dates in, to show when the contribution had happened.

Source/cairo/WaylandCairoSurface.m
```diff
@@ -36,43 +39,45 @@
...
-    static const char template[] = "/weston-shared-XXXXXX";
-    const char *path;
-    char *name;
-    int fd;
-
-    path = getenv("XDG_RUNTIME_DIR");
-    if (!path) {
-	errno = ENOENT;
-	return -1;
+  static const char template[] = "/gnusteo-shared-XXXXXX";
```
fredkiefer, line 48: Is this a typo or intentional? Most likely it should be "gnustep"

fredkiefer (approved):
The formatting looks a lot better now.
It would be great if you could re-add the original dates and maybe fix the typo. After that merging is fine for me.

Headers/wayland/WaylandServer.h
nongio: 👍

Source/cairo/WaylandCairoSurface.m
nongio: :) indeed, I am going to amend the previous commit

Source/cairo/WaylandCairoSurface.m
nongio: it might be superfluous this code, i need to debug it and add a comment for this edge case, the flag used elsewhere is `window->buffer_needs_attach`

ivucica:
nit: That should likely be November 2021, not November 2016.

sent from phone

Source/cairo/WaylandCairoSurface.m
nongio: @fredkiefer i added a comment in the code to explain better when this is needed:
usually the attach() is handled in the configure callback, when resizing a window already configured then the new buffer created should be attached immediately. maybe we could move the attach outside of the surface class...

Source/wayland/WaylandServer+Cursor.m
nongio: I think that the current code structure provides a good separation of concerns: Cursors, Keyboard, Layer/Xdg protocols, Seat and Output. Starting from this we could start collecting feedback on the separate areas and tackle them 1 by 1

fredkiefer: I don't see how your reply is in any way related to my comment. You probably wanted to explain that this code is not yours, you just moved it into a separate file.

nongio: yes, the comment is more related to the general feedback that the PR is very large to be reviewed. Anyway yes, the Cursor code has only been moved in a new file

gcasa:
Merging this on behalf of @nongio since he doesn't have merge permission.  I spoke with him via email.

PR #34, opened 2021-12-14
Wayland shm surface

nongio:
This PR is a refactor of the WaylandCairoSurface, which is an implementation of a Wayland shared memory surface. It includes:
- some documentation on how it works
- a buffer release listener to fix the memory leak present before
- rename to a more specific name as more Wayland surface implementations are possible
- move `wl_surface_attach` calls from the server into the surface class


fredkiefer (approved):
Looks good and only affects the Wayland backend which is still experimental.

PR #38, opened 2022-01-21
Cairo: Disable font hinting when GSScaleFactor is not 1 to fix text d…

anthonyc-r:
…isplay issues

This commit disables font hinting in the cairo backend if GSScaleFactor is not 1. It fixes an issue I posted under libs-gui here: https://github.com/gnustep/libs-gui/issues/134.

### DIFF
--- a/Source/cairo/CairoFontInfo.m
+++ b/Source/cairo/CairoFontInfo.m
@@ -48,6 +48,7 @@ void set_font_options(cairo_font_options_t *options)
   cairo_hint_metrics_t metrics = CAIRO_HINT_METRICS_ON;
   cairo_hint_style_t   style = CAIRO_HINT_STYLE_NONE;
   int hinting = [ud integerForKey: @"GSFontHinting"];
+  float scale = [ud floatForKey: @"GSScaleFactor"];
   int subpixel = 0;
 
   if (hinting == 0) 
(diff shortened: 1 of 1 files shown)


Source/cairo/CairoFontInfo.m
```diff
@@ -48,56 +48,73 @@ void set_font_options(cairo_font_options_t *options)
...
-          hinting = 33;
+        case 0:
+          metrics = CAIRO_HINT_METRICS_DEFAULT;
+          break;
+        case 1:
+          metrics = CAIRO_HINT_METRICS_OFF;
+          break;
+        case 2:
+          metrics = CAIRO_HINT_METRICS_ON;
+NSLog(@"Hinting on");
```
fredkiefer, line 86: Looks like you have  a leftover log statement here.

Source/cairo/CairoFontInfo.m
```diff
@@ -48,56 +48,73 @@ void set_font_options(cairo_font_options_t *options)
...
-      break;
-    case 3:
-      style = CAIRO_HINT_STYLE_MEDIUM;
-      break;
-    case 4:
-      style = CAIRO_HINT_STYLE_FULL;
-      break;
     }
-  
+
```
fredkiefer, line 114: Why all these empty lines?

fredkiefer:
I think the code is OK now. I am still unsure wether we need this change. When somebody manually changes the scaling factor she could as well switch off the font hinting by setting the other environment value to 33, which also is the default value.
Maybe we should document this better?

anthonyc-r:
Setting this to '17' appears to disable font hinting completely, perhaps a better patch would just set hinting to 17 if GSScaleFactor is set to anything other than 1, and GSFontHinting is not explicitly set. I see what you mean about being able to just explicitly set this, however I think that it shouldn't be possible to 'break' rendering by just setting scaling alone... 

I don't know how difficult a change it would be, but perhaps respecting the advancements list (in GSShowGlyphsWithAdvances) would also allow for font hinting and scaling to be set together?

anthonyc-r:
I've updated my commit to still allow for font hinting to be set via GSFontHinting.

fredkiefer:
I really like the new code. Thank you for the patch.

PR #42, opened 2023-09-29
Add headless backend

qmfrederik:
This is another attempt to add a headless backend to GNUstep.  The goal is to be able to run code which interacts with libs-gui in environments where no display is available.  Think legacy applications running in Docker containers and similar environments.

This PR takes a minimalistic approach, and adds an implementation of `GSContext` and `GSDisplayServer` which overrides just enough methods for the headless server to support the workloads I've tested against.

As part of this PR, I've also had to run `autoreconf`.  That generated some noise because I used a newer version of autoreconf.  That noise is in 1ec7eb6ecbac4c8a97a1c870c1dfeecbfc273bc4, for which I can generate a separate PR if that makes it easier to review.


qmfrederik:
@fredkiefer Let me know if this approach works for you.  I'd be happy to make changes to this PR based on your feedback
/cc @gcasa

Headers/headless/HeadlessContext.h
```diff
@@ -0,0 +1,28 @@
+/* HeadlessContext.c - Headless Context
+ * 
+ *  Copyright (c) 2023 Keysight Technologies
```
fredkiefer, line 3: I am a bit confused by this copyright line. Do we really want to keep this especially as you removed every content from the original files?

Overall I think you should go through the file headers once more and adjust these as needed..

Source/headless/HeadlessFontInfo.m
```diff
@@ -0,0 +1,86 @@
...
+- (id) initWithFontName: (NSString *)name
+		 matrix: (const CGFloat *)fmatrix
+	     screenFont: (BOOL)p_screenFont
+{
+  self = [super init];
+  if (!self)
+    return nil;
+
+#ifndef _MSC_VER
+  // Accessing instance variables across module boundaries is not supported by the Visual Studio
```
fredkiefer, line 45: If this is a real issue, we need a better solution for it. Maybe implement a simple version of this method on GSFontInfo in gui and call that form back. But this could be resolved independently of you pull request.

fredkiefer:
Apart from the copyright this looks great. And if we ever need to create PDFs from a headless system we just need to support the headless/cairo combination, similar to cairo in win32.

Source/headless/HeadlessFontInfo.m
qmfrederik: Yes, the proper fix for this is likely getter/setter methods for a bunch of instance variables in libs-gui.  Getting the Win32 backend to compile on msvc will require similar kind of work.  I hope to get to that one day, but it will be on a separate PR ;-)

qmfrederik:
Thanks, @fredkiefer.  I went over the headers for files and streamlined their formatting (and reference the Lesser GPL instead of the Library GPL).  I hope the current format is acceptable to you; let me know if you require any changes.

fredkiefer:
We will have to wait for a decision by @gcasa here. I don't know whether we want to accept code that is dual copyrighted. Most likely this is not valid in most countries. In this specific case it is also quite trivial code maybe it would be possible for you to remove the Keysight copyright?

gcasa:
TestPlant, and therefore Keysight as its successor, signed a corporate copyright disclaimer this should cover anyone that contributes to GNUstep from Keysight.  

Proof of this can be obtained from the FSF and it should be on the disclaimers page on the website (I will share that address in a subsequent reply). 

That being said @qmfrederik i don't think the "Copyright Keysight" notice is needed as I believe simply putting in your name and keysight email address as the author is sufficient.

gcasa:
Here is the relevant page on the site...

https://gnustep.github.io/developers/copyright.html

It's near the end.  I believe I have the email from the copyright clerk in my gmail account as well.

qmfrederik:
I updated the copyright statements to be in line with those found at https://github.com/keysight-eggplant/libs-back. I hope that helps.

fredkiefer (approved):
Now it looks good to me. If there is nothing withstanding from @gcasa it is fine to merge.

gcasa:
@fredkiefer @qmfrederik all good.

PR #46, opened 2023-12-20
try to handle grouping of windows differently

rmottola:
XFCE windowmanager groups in the task bar windows when there are many for one application. What I found is that with two or more GNUstep applications, everything was put into a "GNUstep" item, different from the expected experience
After some trial-and error using xprops on various apps (internal XFCe,, GTK tools, Firefox...) I found out the issue is the WM_CLASS atoms, which consists of two strings

for most application they were set like this:
"app name", "App Name"

for some others, like this:` "org.gnome.appname", "AppName"`

or even a functional name, e.g Firefox has: "Navigator", "Firefox" which I bet is a heritage of when there was Navigator, Mail, Composer, Chat

we were setting `"AppName", "GNUstep"`

Since in the place I only have NSProcessInfo, (using the domain or a functional name like Firefox).. so now we can set e.g.` "Ink", "Ink"`.
At a first test, it works in XFCE!

I wonder:

- [ ] Is/Was "GNUstep" as class used somewhere or for other WMs?
- [ ] Do other places where res_class is set to GNUstep be also changed?

### DIFF
--- a/Source/x11/XGServerWindow.m
+++ b/Source/x11/XGServerWindow.m
@@ -1,3 +1,4 @@
+
 /* XGServerWindows - methods for window/screen handling
 
    Copyright (C) 1999-2020 Free Software Foundation, Inc.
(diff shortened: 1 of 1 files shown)


2xsaiko:
How about setting it to the value of CFBundleIdentifier? That seems to be the closest to "org.gnome.appname".

rmottola:
> How about setting it to the value of CFBundleIdentifier? That seems to be the closest to "org.gnome.appname".

That detail seems only to be done for that GNOME app, you may try to inspect random apps you are running.

Furthermore, i don't know how to access in an efficient way the CFBundleIdentifier in XGServerWindow

rmottola:
@fredkiefer do we use res_class for anything else?

We left this stale... would be nice to make it before release.

rmottola:
@rfm I cleaned up and use the process name, as suggested. The Bundle Identifier was too verbose and unexpected in most cases, except certain GNOME applications, but we like to be more user friendly. I hope the C handling of string and memory is correct? No Copy needed! I copied it from the root window.

PR #48, opened 2024-08-28
Fix build errors with Wayland on Debian 13 (trixie) using Clang 16

pkgdemon:
This pull request resolves several build errors encountered when compiling with Wayland support (./configure --enable-server=wayland) on Debian 13 Testing (codename trixie) using Clang version 16.0.6.

The updates include:

- Selector Naming Issues: Corrected method declarations to properly align with Objective-C selector conventions.
- Function Pointer Compatibility: Fixed mismatched function pointer types to align with expected signatures.
- Implicit Declarations: Added necessary includes and declarations for Wayland functions to prevent implicit declaration errors.
- Type Corrections: Adjusted type usage to match expected formats and avoid type mismatch errors.

If this PR is accepted, I will work on a follow-up PR will address additional build issues on Arch Linux with Clang 18.1.8.

### DIFF
--- a/Source/wayland/WaylandServer+Cursor.m
+++ b/Source/wayland/WaylandServer+Cursor.m
@@ -38,6 +38,9 @@
 #include <linux/input.h>
 #include "wayland-cursor.h"
 
+// Added extern declaration to resolve implicit function declaration error
+extern void wl_cursor_destroy(struct wl_cursor *cursor);
+
 // XXX should this be configurable by the user?
 #define DOUBLECLICK_DELAY 300
 #define DOUBLECLICK_MOVE_THREASHOLD 3
@@ -290,19 +293,17 @@ pointer_handle_button(void *data, struct wl_pointer *pointer, uint32_t serial,
           NSWindow *nswindow = GSWindowWithNumber(window->window_id);
           if (nswindow != nil)
             {
-              // Removed the usage of _windowView as it is not declared
-              // GSStandardWindowDecorationView * wd = [nswindow _windowView];
+              GSStandardWindowDecorationView * wd = [nswindow _windowView];
 
-              // Assuming some alternative logic instead
-              if ([nswindow pointInTitleBarRect:eventLocation])
+              if ([wd pointInTitleBarRect:eventLocation])
                 {
                   xdg_toplevel_move(window->toplevel, wlconfig->seat, serial);
                   window->moving = YES;
                   return;
                 }
-              if ([nswindow pointInResizeBarRect:eventLocation])
+              if ([wd pointInResizeBarRect:eventLocation])
                 {
-                  GSResizeEdgeMode mode = [nswindow resizeModeForPoint:eventLocation];
+                  GSResizeEdgeMode mode = [wd resizeModeForPoint:eventLocation];
 
                   uint32_t edges = XDG_TOPLEVEL_RESIZE_EDGE_BOTTOM_RIGHT;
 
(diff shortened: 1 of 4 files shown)


Source/wayland/WaylandServer+Xdgshell.m
```diff
@@ -28,10 +28,14 @@
 #include "wayland/WaylandServer.h"
 #include <AppKit/NSEvent.h>
 #include <AppKit/NSApplication.h>
+#include <AppKit/NSWindow.h>
+
+// Ensure the function GSWindowWithNumber is declared or included from the appropriate header
+NSWindow *GSWindowWithNumber(int window_id);
```
fredkiefer, line 34: Looks like we have this method in NSGraphics.h, the implementation is in NSWindow.m. But if you want to be more abstract you could also use NSConvertWindowNumberToGlobal.

Source/wayland/WaylandServer+Cursor.m
```diff
@@ -796,7 +799,7 @@ - (void)freecursor:(void *)cid
 {
   // the cursor should be deallocated
   struct cursor * c = cid;
-  wl_cursor_destroy(c->cursor);
+  wl_cursor_destroy(c->cursor); // Ensure wl_cursor_destroy is declared and defined
```
fredkiefer, line 802: You may now remove this comment here and maybe even the one on line 41.

Source/wayland/WaylandServer.m
```diff
@@ -781,7 +781,7 @@ - (void)createSurfaceShell:(struct window *)window
       break;
     case NSPopUpMenuWindowLevel:
       NSDebugLog(@"[%d] NSPopUpMenuWindowLevel", win);
-      [self createPopup:win];
+      [self createPopup:get_window_with_id(wlconfig, win)];
```
fredkiefer, line 784: Why don't you use `window` here?

fredkiefer (commented):
Looks good to me. Could you please check the few comments I left, then it's fine to merge this.

Source/wayland/WaylandServer.m
pkgdemon: This is fixed now.  Sorry I missed this.

Source/wayland/WaylandServer+Cursor.m
pkgdemon: Comments have been removed including line 41.

Source/wayland/WaylandServer+Xdgshell.m
fredkiefer: You could add `#include <AppKit/NSGraphics.h>` here instead.

fredkiefer (approved):

Source/wayland/WaylandServer+Xdgshell.m
pkgdemon: Fix pushed.  Thanks @fredkiefer

pkgdemon: I just realized I probably don't need `NSWindow.h` anymore.  I will test and push another update if it is not needed as well.

pkgdemon: Okay the `NSWindow.h` include was removed.  I think we should be good now.

pkgdemon:
Also just wanted to leave a note to correct a statement made in my description.  The build is working now with Arch Linux and clang 18.  No additional PR will be needed for that.

Issue #49, opened 2024-09-12
on FreeBSD 13.3 issues with extended chars


rmottola:
Small sum up of email thread.

After FreeBSD 13.3 update, extended character typing (e.g. àè or ° and €) started not working and outputting multiple characters.
It appears to be linked to these env. variables set:
`LANG=C.UTF-8`
`XMODIFIERS="@im=kinput2" `

rmottola:
Even without setting XMODIFIERS (when things appear to "work" for the user) I noticed that KeyPress and KeyRelease characters are different for extended characters:
```
2024-09-11 23:19:51.042 Ink[2352:100713] XKeyEvent keyCode 22
2024-09-11 23:19:51.042 Ink[2352:100713] unicode modifiers: 0
2024-09-11 23:19:51.042 Ink[2352:100713] unicode no modifiers: 0
2024-09-11 23:19:51.042 Ink[2352:100713] process_key_event: keys:è ukeys:è
2024-09-11 23:19:51.125 Ink[2352:100713] XKeyEvent keyCode 22
2024-09-11 23:19:51.125 Ink[2352:100713] unicode modifiers: 0
2024-09-11 23:19:51.125 Ink[2352:100713] unicode no modifiers: 0
2024-09-11 23:19:51.125 Ink[2352:100713] process_key_event: keys:Ã¨ ukeys:Ã¨ 
```

It is partially expected: On KeyPress the character conversion is done by using `Xutf8LookupString`  and initialising NSString with `NSUTF8StringEncoding`, while on KeyRelease Xutf8LookupString is not available per X11 spec, then `XLookupString` is used and string is initialised using `NSISOLatin1StringEncoding`. One would expect that at least for Latin1  characters it should be equivalent, but not and the string contains "garbage".

rmottola:
I found out that when the Latin1 is not interpreted correctly, count my XLookupString is > 1, it means it cannot be a single char of a key. It may be 2 long for e.g. à, but 3 for €.

I tried some smartness and if I interpret it as UTF8, all appears to work!!

PR #50, opened 2024-09-12
try to fix extended char handling in NSEvent

rmottola:
Issue described in https://github.com/gnustep/libs-back/issues/49

I found out that sometimes XLookupString returns more than one char for a single key, which does not make much sense for Latin1.
So I suspect I am getting back UTF-8 in that case and indeed, interpreting that string as such works!

Looks like X11 got "extended" here beyond the Spec that says Latin1 only. The question remains why it is happening for characters contained in Latin1?
Last found is that €, not contained in Latin1 gets not back.

However I think this "hack" is pretty safe, since if only one char is returned, Latin1 is used as per-spec.

### DIFF
--- a/Source/x11/XIMInputServer.m
+++ b/Source/x11/XIMInputServer.m
@@ -151,11 +151,17 @@
     {
       /* Always returns a Latin-1 string according to the manpage */
       count = XLookupString (event, buf, BUF_LEN, &keysym, NULL);
-      if (count)
+      if (count == 1)
 	{
 	  keys = [[[NSString alloc] initWithBytes: buf
 					   length: count
 					 encoding: NSISOLatin1StringEncoding] autorelease];
+	}
+      else if (count > 1) // manpage lies and we suppose UTF-8
+	{
+	  keys = [[[NSString alloc] initWithBytes: buf
+					   length: count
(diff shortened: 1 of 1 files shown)


rmottola:
Tried on Linux where I had no issues and continue having no issues with Italian keyboard.

Would you mind testing e.g. with German keyboard?

fredkiefer (approved):
I wanted to test, whether we could switch to NSUTF8StringEncoding even when there is just one character. But as I didn't find the time to do so, it is probably better just to merge this change.

rmottola:
@fredkiefer if you want somehow create a TODO and work with this on the next hacking session together or if I should test a patch for you, I'm available! In the meanwhile we see if this change creates issues.

rfm:
I think the change is good. If the data is utf-8 a single byte is guaranteed to be correct latin1.  If the data is latin1, it's either got the top bit set (and would be invalid utf-8) or it hasn't (and it's the correct ASCII value).  So it seems very safe to me.

rmottola:
@rfm thanks for you opinion, I was a bit wary about this because it is somehow against documentation of X11 and I saw 2 or 3 bytes and was attempted at trial-and-error, but it is safe for 1byte.

PR #52, opened 2024-09-17
Define WINBOOL on non-MinGW platforms

qmfrederik:
The Windows SDK declares `BOOL` as an `int`.  Objective C defines `BOOL` as a `char`. Those two types clash.  MinGW's implementation of the Windows SDK uses the `WINBOOL` type to avoid this clash.  When compiling natively on Windows, we need to manually define `WINBOOL`.
MinGW will define `_DEF_WINBOOL_` if it has defined `WINBOOL` so we can use the same trick here.
See https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/include/ntdef.h#L355

### DIFF
--- a/Source/win32/WIN32Server.m
+++ b/Source/win32/WIN32Server.m
@@ -63,6 +63,18 @@
 
 #include <math.h>
 
+// The Windows SDK declares BOOL as an int.  Objective C defines BOOl as a char.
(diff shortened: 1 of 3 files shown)


qmfrederik:
This could also go in a separate header file; open for suggestions.

fredkiefer (approved):
It looks strange to have this scattered in three files but out of the top of my head I cannot come up with a better place.

PR #53, opened 2024-11-02
Enable default interpolation due to pixelated screen drawing

qmfrederik:
Without this change, images in GUI applications would appear pixelated; with this change, images are rendered much more smoothly.

git-svn-id: svn+ssh://svn.gna.org/svn/gnustep/libs/back/branches/gnustep_testplant_branch@40163 72102866-910b-0410-8b05-ffd578937521


fredkiefer (approved):
I would prefer to have a space after the colon in method calls and no brackets in the return statement. But apart from coding style this change look correct.

qmfrederik:
@fredkiefer Thanks, I made the changes you requested.

PR #57, opened 2025-06-11
Fix close button for wayland

pkgdemon:


fredkiefer:
I don't understand, why this change is needed. Is it because of the resize code just a bit below? If so, we rather have this new code included there. The close button should just work as any other button.
If we really need this code or something similar, could you please make sure to add a space after each colon in a method call? Thank you.

pkgdemon:
Thanks Fred.  Yes I will make the changes.  With the Wayland server this code allows the close button to function when building libs-back with Wayland configured.  The close button does not function with out it when launching GNUstep apps in a Wayland compositor.  Minimize is also broken which I have code to fix also but I was going to follow up with another PR after seeing what changes would be needed for this one.

pkgdemon:
Hi @fredkiefer take a look now when you can.  I think you will like this much better as I do.  I have confirmed it still works to close the window.

fredkiefer:
Thank you for adding the spaces after the colons.
Both this change still looks odd to me. We should not need to have any special code here. To me this looks like the code for resizing is eating up the button pressed event. If that is the case, we should just exclude the button frames from being passed on as resize events.

pkgdemon:
> Thank you for adding the spaces after the colons. Both this change still looks odd to me. We should not need to have any special code here. To me this looks like the code for resizing is eating up the button pressed event. If that is the case, we should just exclude the button frames from being passed on as resize events.

I suspect you may be right.  Let me look into this further.  Thanks.

pkgdemon:
I think it may take some time before I can get back to this fully.  I've kind of lost interest in wayland at the moment and I'm deeper into dbuskit things.  We can close for now if you like @fredkiefer and I can reopen in the future if you want.

PR #59, opened 2025-07-28
wayland: Notes for future readers on use of get_window_with_id and window->terminated.

ivucica:
wayland: Notes for future readers on use of get_window_with_id and window->terminated.

@fredkiefer as maintainer (and perhaps @nongio + @slp as code authors):

Can you please take a closer look at what might be happening here? I don't have a wayland setup to actually build and test this, so I just tried to follow the logic. An automated tool wrote a little note that this `free(window)` might cause a free-after-use, but I don't think it can (even though that is absolutely not clear when I just try to read the code).

Separately, `get_window_with_id` can return a NULL, yet most callsites seem to pretend it can't. I suspect it actually won't be returning NULL in practice, because the window that's gone from local structures will not even be told about -- unless Wayland is similarly asynchronous as X11. Or if the calling functions have bugs.

So, perhaps callers of `get_window_with_id` _should_ deal with it returning NULL. What do you think?

### DIFF
--- a/Headers/wayland/WaylandServer.h
+++ b/Headers/wayland/WaylandServer.h
@@ -169,6 +169,13 @@ struct window
   CairoSurface		       *wcs;
 };
 
+/* get_window_with_id returns the known window with the passed ID, or NULL.
+ *
+ * NULL is returned when the passed window is not known; for example, the
+ * window has been '->terminated' and was removed from the window_list already,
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):

nongio:
@ivucica indeed `get_window_with_id` can return NULL, being a lookup by an integer ID
thanks for updating the documentation

PR #60, opened 2025-10-02
Add FCFontAssetInstaller to libs-back...

gcasa:
Add FCFontAssetInstaller to allow for a more backend sensitive installation of the font.

### DIFF
--- a/Source/fontconfig/FCFontAssetInstaller.m
+++ b/Source/fontconfig/FCFontAssetInstaller.m
@@ -244,8 +244,8 @@
       if ([osName isEqualToString: @"NSWindowsNTOperatingSystem"]
 	  || [osName isEqualToString: @"NSWindows95OperatingSystem"])
 	{
-	  NSString *winDir = [NSString stringWithUTF8String: getenv("USERPROFILE")];
-	  fontsPath = [winDir stringByAppendingPathComponent: @"Fonts"];
+	  NSString *dir = [NSString stringWithUTF8String: getenv("USERPROFILE")];
+	  fontsPath = [dir stringByAppendingPathComponent: @"Fonts"];
 	}
       else if ([osName isEqualToString: @"NSMACHOperatingSystem"])
 	{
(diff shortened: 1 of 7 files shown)


Source/fontconfig/FCFontAssetInstaller.m
```diff
@@ -0,0 +1,296 @@
...
+
+  // Notify system of new font (platform-specific)
+#ifdef __APPLE__
+  // On macOS, fonts are automatically detected when placed in font directories
+  NSLog(@"Font installed to: %@", destinationPath);
+#else
+  NS_DURING
+    {
+      NSTask *task = [[NSTask alloc] init];
+      [task setLaunchPath: @"/usr/bin/fc-cache"];
```
fredkiefer, line 209: I don't like this hard coded path to the executable. Are you sure, it will always be in that directory?

fredkiefer (commented):
Looks good to me, just a few small issues.

Source/fontconfig/FCFontAssetInstaller.m
```diff
@@ -0,0 +1,296 @@
...
+	  fontsPath = [winDir stringByAppendingPathComponent: @"Fonts"];
+	}
+      else if ([osName isEqualToString: @"NSMACHOperatingSystem"])
+	{
+	  fontsPath = [homeDir stringByAppendingPathComponent: @"Library"];
+	  fontsPath = [fontsPath stringByAppendingPathComponent: @"Fonts"];
+	}
+      else // Assume linux/unix
+	{
+	  fontsPath = [homeDir stringByAppendingPathComponent: @".fonts"];
```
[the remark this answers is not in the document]
gcasa: Okay, I will change it to use that.

Source/fontconfig/FCFontAssetInstaller.m
gcasa: I'm using just fc-cache now, it needs to be in the path.

PR #69, opened 2026-01-27
Fix gradient crash on non-RGB NSColor and add regression test

probonopd:
- Fixes an uncaught NSInternalInconsistencyException when gradients include non-RGB colors (e.g., pattern or named color lists) by normalizing colors to an RGB-compatible color space before accessing components in CairoGState (libs-back).
- Adds a regression test `libs-back/Tests/gradient/gradient_test.m` that creates a non-RGB color (pattern) and draws an `NSGradient` into an image with `lockFocus` to ensure no exception is raised.

Testing:
- Built and installed `libs-back` and verified the assistant no longer throws the "Called redComponent on non-RGB colour" exception in UI runs.
- The regression test is included and can be built via `gmake -C libs-back/Tests/gradient` (or by running its tool target) — it draws a gradient into an offscreen image and exits non-zero on failure.


Source/cairo/CairoGState.m
```diff
@@ -1584,10 +1584,26 @@ - (void) drawGradient: (NSGradient*)gradient
...
+          /* Ensure color is in an RGB-compatible colorspace to avoid
+             -redComponent/-greenComponent/-blueComponent raising
+             for non-RGB colors (patterns, color lists, etc.). Try
+             device RGB first, then calibrated RGB, and finally
+             fall back to a safe opaque black if conversion fails. */
+          NSColor *rgbColor = [color colorUsingColorSpaceName: NSDeviceRGBColorSpace];
+          if (!rgbColor) rgbColor = [color colorUsingColorSpaceName: NSCalibratedRGBColorSpace];
+          if (!rgbColor) rgbColor = [NSColor colorWithCalibratedWhite:0.0 alpha:1.0];
+
+          @try {
```
fredkiefer: Could you please use the Marcos here to allow this code to work even with very old gcc? This is not strictly required but is part of the GNUstep coding style.

fredkiefer (changes requested):
Could you please rebase this PR. That way the tests will now be executed by the pipeline.
You should also rewrite the test directory to work together with our test framework. Add a Testinfo file and move the make instructions into a preamble file

probonopd:
@fredkiefer thanks for your review. I hope I addressed it correctly; if not, please let me know. Thanks!

Source/cairo/CairoGState.m
```diff
@@ -1584,10 +1584,26 @@ - (void) drawGradient: (NSGradient*)gradient
...
-          green = [color greenComponent];
-          blue = [color blueComponent];
-          alpha = [color alphaComponent];
+          /* Ensure color is in an RGB-compatible colorspace to avoid
+             -redComponent/-greenComponent/-blueComponent raising
+             for non-RGB colors (patterns, color lists, etc.). */
+          NS_DURING
+            {
+              NSColor *rgbColor = [color colorUsingColorSpaceName: NSDeviceRGBColorSpace];
+              if (!rgbColor) rgbColor = [color colorUsingColorSpaceName: NSCalibratedRGBColorSpace];
```
fredkiefer, line 1593: I am not sure if there are any cases where these fallbacks would succeed when the first call fails. They don't do any harm, so I leave them in.

Tests/cairo/gradient.m
```diff
@@ -0,0 +1,82 @@
...
+  [NSApplication sharedApplication];
+
+  /* Create a small pattern image and make a pattern color (non-RGB). */
+  NSImage *pat = [[NSImage alloc] initWithSize: NSMakeSize(4, 4)];
+  [pat lockFocus];
+  [[NSColor colorWithCalibratedRed: 1.0 green: 0.0 blue: 0.0 alpha: 1.0] setFill];
+  NSRectFill(NSMakeRect(0, 0, 4, 4));
+  [pat unlockFocus];
+  NSColor *pattern = [NSColor colorWithPatternImage: pat];
+  [pat release];
```
fredkiefer, line 37: Could you please switch to using the Macros? We are not that strict for test code, still it is the preferred way.

Tests/cairo/gradient.m
```diff
@@ -0,0 +1,82 @@
...
+  NSColor *white = [NSColor colorWithCalibratedWhite: 1.0 alpha: 1.0];
+
+  /* Linear gradient: drawInRect:angle: calls drawGradient:fromPoint:toPoint:
+   * (linear path).  Without the fix, non-RGB stops are silently skipped. */
+  NSGradient *lg = [[NSGradient alloc]
+                     initWithStartingColor: pattern
+                               endingColor: white];
+  NSImage *limg = [[NSImage alloc] initWithSize: NSMakeSize(100, 100)];
+  [limg lockFocus];
+  PASS_RUNS(([lg drawInRect: NSMakeRect(0, 0, 100, 100) angle: 90.0]),
```
fredkiefer, line 48: Not raising an exception is a rather raw test. Please have a look at the tests by DTW-Thalion to see how to check the drawn data.

fredkiefer (approved):
I realy would like to see more changes here, but I will merge anyway. Feel free to provide these changes later on.

PR #77, opened 2026-04-23
 Opal backend: make functional again

pkgdemon:
## Summary                                                                                                                                             
  
  The opal backend in libs-back has been non-functional for years.                                                                                       
  This PR is the minimum set of changes to make it usable again.  
  All changes are confined to `Source/opal/` plus one declaration in
  `Headers/opal/OpalSurface.h`.     

This does require my other PR to be merged:

https://github.com/gnustep/libs-opal/pull/16                                                                                                                     
                                          
  ## What's fixed                                                                                                                                        
                                                                                                                                                         
  - **Surface attachment.** `OpalSurface` no longer calls                                                                                                
    `OPX11ContextCreate` before the X window is mapped (which produced                                                                                   
    a non-drawing context). A new lazy `-ensureX11Context` creates the
    X11 context on first expose; the accessor calls it so every                                                                                          
    `OpalGState.m` consumer is covered without per-site edits. Resize
    is handled for free since `XGServer` recreates the surface.                                                                                          
  - **Drawing ops** (`OpalGState.m`): filled in missing paths, fixed                                                                                     
    CTM and clipping around the compositing path.                                                                                                        
  - **Font hinting** (`OpalFontInfo.m`): propagate hint settings to                                                                                      
    cairo font options; correct `_cairo_extents_for_NSGlyph`.                                                                                            
  - **Gradients and blend modes**: implemented CG linear/radial                                                                                          
    gradients; mapped NSCompositing operators onto CG blend modes.                                                                                       
  - **Grayscale colors**: convert `NSColor` gray values to device-RGB                                                                                    
    `CGColorRef` so gray fills/strokes take effect.                                                                                                      
  - **Window-drag smearing**: flush both backing and X11 contexts in                                                                                     
    `-handleExposeRect:`.                                                                                                                                
                                                                                                                                                         
  ## Scope                                                        
                                                                                                                                                         
  No other backends touched. No shared code (`Source/gsc/`,       
  `Source/x11/`, `Source/gui/`) touched. Cairo path is unaffected.                                                                                       
                                                                  
  ## Building and testing                                                                                                                                 
                                                                  
  ```sh                                                                                                                                                  
  cd libs-back                                                                                                                                                                                                                                                                         
  ./configure --enable-graphics=opal --enable-server=x11                                                                                                 
  make -j$(nproc)                                                 
  sudo -E make install
```                        
                                          
 Verified: opal-backed apps render correctly. Workspace (cairo-backed)

<img width="100%" alt="opal_proof3" src="https://github.com/user-attachments/assets/b57047b5-cbec-4c9a-86f4-6f6a9c7f658c" />

### DIFF
--- a/Source/opal/OpalContext.m
+++ b/Source/opal/OpalContext.m
@@ -45,6 +45,7 @@
 
 + (void) initializeBackend
 {
+  NSLog(@"OpalContext: initializeBackend called");
   [NSGraphicsContext setDefaultContextClass: self];
 
   [GSFontEnumerator setDefaultClass: [OpalFontEnumerator class]];
@@ -63,16 +64,8 @@
 
 - (BOOL) isDrawingToScreen
 {
-#warning isDrawingToScreen returning NO to fix DPSimage
-  return NO;
-
-  // NOTE: This was returning NO because it was not looking at the
-  // return value of GSCurrentSurface. Now it returns YES, which
-  // seems to have broken image drawing (yellow rectangles are drawn instead)
   OpalSurface *surface;
-
   [OGSTATE GSCurrentSurface: &surface : NULL : NULL];
-
   return [surface isDrawingToScreen];
 }
 
@@ -175,6 +168,7 @@
 }
 
 - (void) GSSetDevice: (void *)device
+
                     : (int)x
                     : (int)y
 {
--- a/Source/opal/OpalContext.m
+++ b/Source/opal/OpalContext.m
@@ -143,17 +143,15 @@
       return NO;
     }
 
-  // FIXME: Allow more image types as soon as the Opal backend handles them correctly
   colorSpaceName = [bitmap colorSpaceName];
-  if (![colorSpaceName isEqualToString: NSDeviceRGBColorSpace] &&
-      ![colorSpaceName isEqualToString: NSCalibratedRGBColorSpace])
-    {
-      return NO;
-    }
-  else
+  if ([colorSpaceName isEqualToString: NSDeviceRGBColorSpace] ||
+      [colorSpaceName isEqualToString: NSCalibratedRGBColorSpace] ||
+      [colorSpaceName isEqualToString: NSDeviceWhiteColorSpace] ||
+      [colorSpaceName isEqualToString: NSCalibratedWhiteColorSpace])
     {
       return YES;
     }
+  return NO;
 }
 
 - (void) GSCurrentDevice: (void **)device : (int *)x : (int *)y
--- a/Source/opal/OpalContext.m
+++ b/Source/opal/OpalContext.m
@@ -166,7 +166,6 @@
 }
 
 - (void) GSSetDevice: (void *)device
-
                     : (int)x
                     : (int)y
 {
--- a/Source/opal/OpalGState.m
+++ b/Source/opal/OpalGState.m
@@ -1104,4 +1104,37 @@ doesn't support to use the receiver cairo target as the source. */
   free(savedClip);
 }
 
+
+// Gradient rendering - stub implementations to prevent crashes
+- (void) drawGradient: (NSGradient*)gradient
+            fromPoint: (NSPoint)startPoint
+              toPoint: (NSPoint)endPoint
+              options: (NSUInteger)options
+{
+  // TODO: Implement using CGGradientCreateWithColorComponents + CGContextDrawLinearGradient
+  NSLog(@"OpalGState: drawGradient (linear) - stub, drawing fallback fill");
+
+  // Fallback: fill with the first color of the gradient
+  CGContextRef ctx = [self CGContext];
+  if (ctx && gradient)
+    {
+      NSColor *color = [gradient interpolatedColorAtLocation: 0.0];
+      CGFloat r, g, b, a;
+      [[color colorUsingColorSpaceName: NSCalibratedRGBColorSpace]
+          getRed: &r green: &g blue: &b alpha: &a];
+      CGContextSetRGBFillColor(ctx, r, g, b, a);
+    }
+}
+
+- (void) drawGradient: (NSGradient*)gradient
+           fromCenter: (NSPoint)startCenter
+               radius: (CGFloat)startRadius
+             toCenter: (NSPoint)endCenter
+               radius: (CGFloat)endRadius
+              options: (NSUInteger)options
+{
+  // TODO: Implement using CGGradientCreateWithColorComponents + CGContextDrawRadialGradient
+  NSLog(@"OpalGState: drawGradient (radial) - stub");
+}
+
 @end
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by pkgdemon** [comment]
@gcasa Here is the last PR for what we talked about last month in gnustep meeting.  ^^

**Comment by pkgdemon** [comment]
@ivucica @fredkiefer Let me know if this is something GNUstep might want or not in the near term.  

There are still some issues with it I need to work out namely terminal rendering, fonts rendering for menus that I've been stuck on but I've made some progress for terminal.  Then there is a delay launching GNUstep apps in non EMWH compliant window managers (that I've fixed in another more experimental fork not here because I don't want to touch code outside of Opal in this PR).  

A few months ago first I started out tinkering with libs-corebase, libs-opal, libs-quartzcore integration as an idea to facilitate easier application porting for things like texmate.  It then extended to an idea for a future display server to replace both X11 and wayland as I learned about this code in libs-back that I've been working on for a few months in parallel.

image

So I have this WindowServer.app piece somewhat working, not pushed anywhere yet.  Ignore the theme issues rendering a double menu and so on (trivial to solve, but is a hack to reparent menus in Menu.app Gershwin uses) I have been working to use some shared components that X or wayland would use such as libinput.

telegram-cloud-photo

Anyway my goals was to fix the terminal, menu items rendering issue before pivoting back to the display server.  But the idea to use a refreshed Opal vs Cairo drawing backend appealed to me for this project as being more architecturally correct over time.  So step 1 is fixing opal to be on par with Cairo for X11 (I consider this PR a great first step with a follow up coming in time).  Step 2 is finish WindowServer.app.  Step 3 is port WindowManager.app to X11.app (think xquartz).  

The ideas is that there would a fully independent display server model for running GNUstep apps natively only requiring KMS drivers, libinput, and a few EGL related packages.  Then I would make a separate application for optional compatibility with X11 applications (mostly I already have it with WIndowManager.app just needs a refactor), and eventually I would make something similar for Wayland applications both as optional bundles.

File: Source/opal/OpalContext.m [hunk 1]
```diff
@@ -175,6 +166,7 @@ - (void) GSCurrentDevice: (void **)device : (int *)x : (int *)y
 }
 
 - (void) GSSetDevice: (void *)device
+
```
fredkiefer: This newline looks wrong.

Source/opal/OpalGState.m
```diff
@@ -734,8 +767,79 @@ - (void) GSSendBezierPath: (NSBezierPath *)newpath
...
+
+  // Draw into temp context
+  CGContextDrawImage(tmpCtx, CGRectMake(0, 0, w, h), subImg);
+  CGImageRelease(subImg);
+
+  // Extract pixel data
+  unsigned char *srcData = (unsigned char *)CGBitmapContextGetData(tmpCtx);
+  if (!srcData) { CGContextRelease(tmpCtx); return nil; }
+
+  // Convert from BGRA premultiplied to RGBA non-premultiplied
```
fredkiefer, line 807: If I remember correctly, there should be a helper method in our implementation of NSBitmapImage to do this.

Source/opal/OpalGState.m
```diff
@@ -1104,4 +1222,126 @@ - (void) restoreClip: (void *)savedClip
...
+      if (rgb)
+        {
+          components[i*4+0] = [rgb redComponent];
+          components[i*4+1] = [rgb greenComponent];
+          components[i*4+2] = [rgb blueComponent];
+          components[i*4+3] = [rgb alphaComponent];
+        }
+      else
+        {
+          // Fallback for non-RGB colors
```
fredkiefer: This seems a bit strange as there are so many different colour spaces, wouldn't it be better to convert directly to RGB?

Source/opal/OpalGState.m
```diff
@@ -1104,4 +1222,126 @@ - (void) restoreClip: (void *)savedClip
...
+           fromCenter: (NSPoint)startCenter
+               radius: (CGFloat)startRadius
+             toCenter: (NSPoint)endCenter
+               radius: (CGFloat)endRadius
+              options: (NSUInteger)options
+{
+  CGContextRef ctx = CGCTX;
+  if (!ctx || !gradient) return;
+
+  NSInteger stops = [gradient numberOfColorStops];
```
fredkiefer: This duplicates the code from the last method. It would be better to have a shared conversion method or function.

fredkiefer (commented):
This looks like a great improvement. The code here is really in a very bd state and I am glad someone is fixing it.
Please feel free to review some of my suggestions but I am willing to merge even without further changes.

Source/opal/OpalGState.m
pkgdemon: Ah thank you for the tip.  I will look into this, and try to get things refactored this week.

Source/opal/OpalContext.m
pkgdemon: Fixed in b535c1da785c936b1e525ecfb9c7f4ed685743c0

Source/opal/OpalGState.m
pkgdemon: @fredkiefer looking over all this a few times it seems the comment was wrong.  It does not actually unpremultiply.  It does this `// Reorder host-order ARGB bytes into RGBA byte order` not `// Convert from BGRA premultiplied to RGBA non-premultiplied`.  

With that said after reviewing what I could find for NSBitmapImage I am not sure it would work here since all of the conversion methods seem to assume samples are in R, G, B order and it seems none swap red and blue:
* https://github.com/gnustep/libs-gui/blob/52083165452f8f281abb627dcdbcd19d4735ce25/Source/NSBitmapImageRep.m

CairoGState.m seems to do the equivalent reorder inline, though it's endian-aware in a way the Opal code currently isn't:
* https://github.com/gnustep/libs-back/blob/master/Source/cairo/CairoGState.m#L935-L954

Source/opal/OpalGState.m
pkgdemon: I believe this is addressed in 45a979472c7f44fa960eb7900166f69765ea1891.  I can confirm there are no regressions after the changes.

Source/opal/OpalGState.m
pkgdemon: I believe this one is addressed here https://github.com/pkgdemon/libs-back/commit/8ea24deba851dfd9d4fd5d1d95e1676da8e1dab1 I can confirm there are no regressions after the changes.

pkgdemon:
> This looks like a great improvement. The code here is really in a very bd state and I am glad someone is fixing it. Please feel free to review some of my suggestions but I am willing to merge even without further changes.

Thanks @fredkiefer I've addressed 3/4 but I am not sure how to address this one yet for 4/4.  Do you think just a comment update is in order for now?  https://github.com/gnustep/libs-back/pull/77#discussion_r3142648115

Source/opal/OpalGState.m
fredkiefer: There is actually a method in NSBitmapImage that does this (- _convertToFormatBitsPerSample:...), but if your implementation is fast enough for your case, we should go with it.

PR #78, opened 2026-05-10
opal: define XRENDER for x11 builds

pkgdemon:
This fixes startup delay for GNUstep apps with Opal by adding the same conditional that Cairo has.  I also updated configure with autoconf, but if GNUstep team would prefer to do that themselves, let me know and I can wipe the commit and force push.


fredkiefer (approved):
I wanted to regenerate configure anyway and will do so later today :-)

PR #80, opened 2026-05-16
Wayland/feature parity m0 m5

DMJC:
This adds support for NSOpenGL on Wayland as well as fixes some issues with drag and drop and Keyboard and Mouse input handling. I'm looking for assistance/guidance on what needs to be done to get this merged. I'm expecting to have to go back and edit this.

For the OpenGL support I had Claude compare the X11/Wayland drivers and implement the Wayland GL support based on the X11 implementation. This is working in the MyGL.app test application as well as other GL applications running on Wayland.

I had ChatGPT run a report against the X11/Wayland drivers and then had Claude implement missing features from Wayland. This fixed Keyboard handling in terminal.app as I can now use tab-autocompletion where it was broken before. It also fixed drag/drop support on Wayland. I can now drag/drop tabs in Firefox properly and It partially fixed drag/drop support in Gorm.app.


wayland_feature_implementation_plan.md
```diff
@@ -0,0 +1,173 @@
+# Wayland Backend Feature Implementation Plan (vs X11 parity)
```
fredkiefer: It is surely helpful for you to have these two files, but we should store them someplace else. Either in the Documentation folder or in the wayland folder itself.

$GNUSTEP_MAKEFILES/config.guess
```diff
@@ -0,0 +1,1815 @@
+#! /bin/sh
```
fredkiefer, line 1: Can you please explain what this and the next file are for and why they live in a folder with a rather strange name? And why this went unnoticed? This could give the impression that you never even looked through the changes your AI did produce. And if you didn't , why should I?
I really would like to see the changes from this PR to be applied but this will require more commitment from your side.

wayland_feature_implementation_plan.md
DMJC: I've removed the automated testing files from the folder.  The Smoke test was specifically used to verify that the changes to text input/feature parity with the X11 backend were working. The planning.MD files were created by Chat GPT when I asked it to compare the X11 backend and the Wayland backend for feature differences. Those plan files are no longer required as the plans were executed by claude and were successful.

Source/GNUmakefile.preamble
```diff
@@ -45,7 +45,7 @@ CONFIG_SYSTEM_INCL += $(GRAPHIC_CFLAGS)
 
 # Additional library directories the linker should search
 #ADDITIONAL_LIB_DIRS +=
-CONFIG_SYSTEM_LIB_DIR += $(GRAPHIC_LFLAGS)
+CONFIG_SYSTEM_LIB_DIR += $(GRAPHIC_LFLAGS) -lGL -lEGL -lwayland-egl
```
fredkiefer, line 48: We should not add libraries here. This should rather be done in the configure script when the libraries are requested and present.

configure.ac
```diff
@@ -26,7 +26,7 @@
 builtin(include, pkg.m4)dnl
 
 AC_INIT
-AC_PREREQ([2.73])
+AC_PREREQ([2.72])
```
fredkiefer, line 29: Why did you go back to the older version here? You otherwise had no changes to this file and so a regeneration of the configure file should not be necessary.

$GNUSTEP_MAKEFILES/config.guess
DMJC: That entire folder is unnecessary. I've removed it from the Source tree. It appears that it was the $GNUSTEP_MAKEFILES folder e.g /usr/GNUstep/System/Library/Makefiles. At some point rather than link/refer to the folder the folder contents were copied into that location. Subsequent runs never used it.

fredkiefer:
In its current form this PR is almost unreviewable. We need to split it up. It looks like you are building up on already existing other changes. Should we try to get this other changes in one by one first?  Are these mergable by themselves or are there changes required?
And after that we still need to split up your changes even further to be able to get them in. You asked for help on the mailing list. Did anybody reply? If not we could try to do this together, but my time is rather limited.

configure.ac
DMJC: I manually regenerated the file and when I was merging it I was getting a mismatch preventing it from merging. I'm on Debian 14 with Autoconf 2.72.

DMJC:
Thanks Fred for taking some time on this, I will try to split the changes up by feature and resubmit the requests. Now that I know some of the problems I'll try to make this easier to merge.

There are still issues in the Wayland backend, but they are mostly around popup/right click menus, and popup windows like file open/save dialogs disappearing. This was the case before I made my changes.

fredkiefer:
> I will try to split the changes up by feature and resubmit the requests. Now that I know some of the problems I'll try to make this easier to merge.

Sounds great. Looking forward to help you with this.

DMJC:
No longer required, closing. Will refactor/resubmit in other pull requests.

PR #81, opened 2026-05-18
OpenGL Support for Wayland

DMJC:
This is a preliminary implementation of OpenGL support for Wayland. Some mouse handling behaviour is broken, which will be addressed with subsequent pull requests.


Source/wayland/WaylandServer+Keyboard.m
```diff
@@ -105,18 +107,59 @@
 keyboard_handle_enter(void *data, struct wl_keyboard *keyboard, uint32_t serial,
 		      struct wl_surface *surface, struct wl_array *keys)
 {
-  // NSDebugLog(@"keyboard_handle_enter");
-  WaylandConfig	*wlconfig = data;
+  NSDebugLog(@"keyboard_handle_enter");
```
fredkiefer: Do you really want this active?

Source/wayland/WaylandServer+Xdgshell.m
```diff
@@ -76,22 +73,11 @@
...
-                             timestamp:0
-                          windowNumber:(int) window->window_id
-                               context:GSCurrentContext()
-                               subtype:GSAppKitWindowFocusIn
-                                 data1:0
-                                 data2:0];
-
-      [nswindow sendEvent:ev];
-    }
+  /* Keyboard focus is now handled exclusively by keyboard_handle_enter/leave,
```
fredkiefer: Should we keep this comment here?

fredkiefer:
This PR are still at least three changes put into one, OpenGL, Inputserver and smaller improvements. Do you see a chance of splitting things up even further? 
You did forget to add the configure.ac file the next time somebody recreates the configure file all your changes there will be lost. 
With your OpenGL changes in place it is no longer possible to compile the Wayland backend when these libraries are not present. I would prefer if they are an optional dependency and the backend would work without them as before.

fredkiefer:
Oops, it is even worse. You added the OpenGL libraries as a hard dependency for all backends. Please have a look at the X11 implementation to see how to include these conditionally.

DMJC:
I've gated the GL support in the Wayland backend and tried to separate the other changes. There is a cursor fix that is included for stability.

fredkiefer (approved):
Looks good enough for me. Let's merge this and go on to the next step and continuously improve this backend.

PR #84, opened 2026-06-17
try to more closely match freedesktop specifications: distinguish bet…

rmottola:
…ween submenus and torn-off menus; attempt to set toolbar instead of dock for the main menu

### DIFF
--- a/Source/x11/XGServerWindow.m
+++ b/Source/x11/XGServerWindow.m
@@ -3445,11 +3445,15 @@ swapColors(unsigned char *image_data, NSBitmapImageRep *rep)
             }
           else if (level == NSMainMenuWindowLevel)
             {
-              data[0] = generic._NET_WM_WINDOW_TYPE_DOCK_ATOM;
+              data[0] = generic._NET_WM_WINDOW_TYPE_TOOLBAR_ATOM;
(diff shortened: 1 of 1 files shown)


rmottola:
I think this branch is quite correct, however we cannot test it until GUI distinguishes (and sets) tear-off and submenus correctly

fredkiefer:
As I wrote in the gui PR, we need to change all the backends not just x11. Also this change needs to be checked on many window managers, especially on the one I am using (KDE). Otherwise we have an improvement for some, but a regression for others.

rmottola:
> As I wrote in the gui PR, we need to change all the backends not just x11. Also this change needs to be checked on many window managers, especially on the one I am using (KDE). Otherwise we have an improvement for some, but a regression for others.

Of course, we need testing. Right now it appears to do no harm on windowmaker (but doesn't fix an annoying layering detail either). On Mac-X11 now all menus are movable, ont what we like, but more consistent. Will test xfce, please test kde.

I think freedesktop adherence is good. Reading the documentation, these styles were added specifically for NeXT style interface (!)
Currently, even with the gui change, submenus and teared-off menus are still the same, I'd like to fix that. Try on your side with props.

PR #86, opened 2026-06-30
Wayland backend: build under modern clang, fix resource leaks and a title comparison

DTW-Thalion:
## Summary

This fixes a build break and three resource/correctness bugs in the Wayland backend, and adds the first regression test directory for libs-back.

### `WaylandCairoShmSurface.m`
- **Build fix (modern clang):** the `@implementation` re-declared the `pbuffer` ivar that the header `@interface` already declares. The non-fragile-ABI clang ObjC compiler rejects this (`instance variable is already declared`), so the Wayland Cairo backend did not build at all on a current toolchain. The redundant `@implementation` declaration is removed (GCC never required it, so this compiles on both).
- **`createShmBuffer` leak:** the `pool_buffer` allocated up front was leaked on every error path — a zero-area request, a failed pool file, or a failed `mmap` — and the `mmap`-failure path also leaked the open `poolfd`. It now frees the buffer (and closes the fd) before returning `NULL`.

### `WaylandServer.m`
- **`-titlewindow::`:** the guard meant to skip the default `"Window"` title compared the `NSString` with `==` (pointer identity), which is essentially never true, so the title was always re-sent. It now uses `-isEqualToString:`.
- **`-dealloc`:** `-_initWaylandContext` allocates a `WaylandConfig` and opens a `wl_display` connection that were never released. `-dealloc` now disconnects the display and frees the config.

### Tests
`Tests/cairo/` is the first test directory for libs-back. `shmbuffer.m` exercises `createShmBuffer`'s zero-area path (which needs no compositor) by compiling the backend source directly against a small `CairoSurface` stand-in.

## Validation (Ubuntu 24.04, clang 18)

- The `back` bundle now compiles and links (it previously failed on the duplicate ivar).
- The new test passes through `gnustep-tests` (`1 Passed test`).
- Built with `-fsanitize=address`, the test fails before the leak fix — `LeakSanitizer: 6400 byte(s) leaked in 100 allocation(s) ... createShmBuffer WaylandCairoShmSurface.m:133` — and is clean after it.
- `-titlewindow::` and `-dealloc` are exercised only with a live Wayland compositor, so they are validated by inspection + the bundle compile rather than the test.


fredkiefer (changes requested):
Most of the changes look good to me. What I am not sure about is the test. That would only work on a Wayland system and I did not see protection for that.
This is the reason, why there are no tests in back, as they can only be for a specific system.

DTW-Thalion:
> Most of the changes look good to me. What I am not sure about is the test. That would only work on a Wayland system and I did not see protection for that. This is the reason, why there are no tests in back, as they can only be for a specific system.

I will take some time and sort that out - I think I should be able to arrange protection across backends.

DTW-Thalion:
Addressed. A test that compiles a backend source can only build for that backend, so this one is now guarded on the backend `configure` actually selected, in two layers:

- **Source** (`shmbuffer.m`): includes `config.h` and only compiles the wayland+cairo source / runs the check under `#if BUILD_SERVER == SERVER_wayland && BUILD_GRAPHICS == GRAPHICS_cairo`; otherwise `main()` simply `SKIP`s the set.
- **Makefile** (`GNUmakefile.preamble`): the cairo/wayland include dirs and `pkg-config` libraries are added only under `ifeq ($(BUILD_SERVER)-$(BUILD_GRAPHICS),wayland-cairo)` (from `config.make`), so no other backend pulls in cairo or wayland at all.

So it builds on every backend and runs only where it applies. Verified on this branch by reconfiguring:

```
wayland + cairo  ->  1 Passed test
x11 + cairo      ->  Skipped set: shmbuffer.m ... back is not built with the wayland+cairo backend
                     1 Skipped set
```

I've written it as a reusable pattern rather than a one-off: guard the source on `config.h`'s `BUILD_SERVER` / `BUILD_GRAPHICS`, guard the dependencies in the preamble on the same via `config.make`, and `SKIP` cleanly otherwise. That should let `back` carry targeted regression tests (like this leak check) without any other backend's `make check` ever seeing a failure.

fredkiefer:
Thank you, now it looks great. What is still missing is a mechanism to start that new test at all. Most likely it is sufficient to add `Tests` as a subproject in the main GNUmakefile.

DTW-Thalion:
Done. I've added a `Tests/GNUmakefile` that drives the gnustep-make test framework (the same thin wrapper libs-base and libs-gui use) and listed `Tests` as a subproject in the top-level `GNUmakefile`, so `make check` now descends into it and runs the backend tests.

Since each backend test guards itself on the backend actually built, `make check` works for any configuration — verified by reconfiguring:

```
wayland + cairo  ->  1 Passed test
x11 + cairo      ->  1 Skipped set   (no failures)
```

So `make check` on other backends never sees a failure from this test; it just skips cleanly.

DTW-Thalion:
I've refined the test wiring so the infrastructure is backend-generic rather than cairo-specific, so other backends (x11, ...) can add regression tests without touching this makefile:

- `Tests/GNUmakefile` discovers every backend test subdirectory (each marked by a `TestInfo` file, named for its `Source/` backend directory) and runs them all. A new backend's tests are picked up automatically.
- The per-directory `GNUmakefile` is generated by the test framework at check time from `GNUmakefile.preamble`, so I've dropped it from the commit — the test dir now carries only its sources (`shmbuffer.m`, `TestInfo`, `GNUmakefile.preamble`), matching the test directories in the other packages.

Each test still guards itself on the backend actually built, so `make check` works for any configuration. Verified:

```
wayland + cairo  ->  1 Passed test
x11 + cairo      ->  1 Skipped set   (no failures)
```

and confirmed that a second backend directory is discovered and run alongside `cairo` when present.

DTW-Thalion:
@fredkiefer as soon as you are satisfied and can merge this in, I have a bunch more tests to submit, but needs this PR merged in so it has the Tests subdirectory and wiring.

DTW-Thalion:
@fredkiefer can you merge this in please.

PR #87, opened 2026-07-09
tests: cover the gsc colour conversions and GSFunction evaluator

DTW-Thalion:
Adds tests for backend-independent code, using the Tests harness from #86.

`Tests/gsc/gscolors.m` and `Tests/gsc/gsfunction.m` cover the colour-space conversions in `Source/gsc/gscolors.c` (gray/rgb/hsb/cmyk and their round trips, under-colour-removal) and the type-0 sampled-function evaluator in `Source/gsc/GSFunction.m` (linear and bilinear interpolation, domain normalisation and input clamping, range and decode scaling, 8- and 16-bit samples, malformed-spec rejection, and a check that the cached `GSFunction2in3out` agrees with the base evaluator).

The gsc sources are plain C/Foundation with no backend dependency, so these compile the source in directly and need no per-backend guard. 48 tests, all passing.

### DIFF
--- a/Tests/gsc/GNUmakefile.preamble
+++ b/Tests/gsc/GNUmakefile.preamble
@@ -1,0 +1,4 @@
+ADDITIONAL_INCLUDE_DIRS += -I$(GNUSTEP_BUILD_DIR)/../../Headers \
+                           -I$(GNUSTEP_BUILD_DIR)/../../Source \
+                           -I../../Headers -I../../Source \
+                           -I$(GNUSTEP_BUILD_DIR)/../.. -I../..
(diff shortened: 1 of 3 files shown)


fredkiefer:
The xdps backend is no longer in active usage, we should not add tests for this code.
Also there are comments in the makefiles that reference specific files. When the tests evolve this comments won't fit any more. Best we don't add them now, that way they won't become obsolete..

DTW-Thalion:
Dropped the xdps tests and removed the file-referencing comments from the preamble.

fredkiefer (approved):

PR #88, opened 2026-07-09
back/gsc: fix gray-to-CMYK inverting the black channel

DTW-Thalion:
`gsColorToCMYK`'s gray case set the black channel K to the gray value itself (`field[3] = field[0]`). A device gray is a brightness (`gsGrayToRGB` maps v to (v,v,v)), so black (gray 0) must give K = 1 and white (gray 1) K = 0, that is K = 1 - v. As written the two ends were swapped, and a gray -> cmyk -> rgb round trip inverted, disagreeing with the gray -> rgb -> cmyk path.

Set `field[3] = 1.0 - field[0]`. `Tests/gsc/cmyk_gray.m` checks the endpoints and the round trip; it fails before the change and passes after.

### DIFF
--- a/Source/gsc/gscolors.c
+++ b/Source/gsc/gscolors.c
@@ -193,7 +193,7 @@ gsColorToCMYK(device_color_t *color)
   switch(color->space)
     {
     case gray_colorspace:
-      color->field[3] = color->field[0];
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):
Thank you for spotting this!

PR #92, opened 2026-07-09
back/x11: do not spin in count_offset for a zero mask

DTW-Thalion:
`count_offset()` shifts a colour mask right until its lowest bit is set, to find the mask's bit offset. With a zero mask the condition `(mask & 1) == 0` never becomes false and `mask` stays zero, so it loops forever. It is called on a TrueColor visual's red/green/blue masks, which are non-zero for a valid visual, but a malformed visual with a zero mask hangs `RCreateContext`.

Stop when the mask reaches zero, returning 0.

`Tests/x11/context.m` covers this: it compiles `context.c` in with its wraster siblings (the same arrangement the convert test in #94 uses) and checks `count_offset` over a range of masks including zero. `count_offset` needs no X server, so the test runs wherever the x11 backend is built. Before this change the zero-mask case loops forever, so a revert shows up as a hang rather than a wrong value.

### DIFF
--- a/Source/x11/context.c
+++ b/Source/x11/context.c
@@ -607,7 +607,7 @@ count_offset(unsigned long mask)
     int i;
     
     i=0;
-    while ((mask & 1)==0) {
+    while (mask != 0 && (mask & 1)==0) {
 	i++;
 	mask = mask >> 1;
     }
--- a/Tests/x11/GNUmakefile.preamble
+++ b/Tests/x11/GNUmakefile.preamble
@@ -1,0 +1,23 @@
+# Extra build settings for the x11 backend tests.
+#
+# A backend test compiles a backend source file directly, so it can only build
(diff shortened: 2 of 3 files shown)


DTW-Thalion:
The x11 CI jobs fail this at the link step (`cannot find -lXmu`): the test compiles `context.c` in to reach `count_offset`, and `context.c` calls `XmuLookupStandardColormap`, so it needs libXmu, which those jobs do not install. Same cause and fix as #94 (add `libxmu-dev` to the x11 jobs); see the note there.

DTW-Thalion:
The x11 CI jobs are red here because they lack libXmu (and a display), not because of this change. #95 adds libXmu/libXext and runs the tests under Xvfb on those jobs; once it lands this needs a rebase onto master to go green.

fredkiefer (approved):

PR #94, opened 2026-07-09
tests: round-trip test for x11 RConvertImage (headless via Xvfb)

DTW-Thalion:
Adds a round-trip test for `RConvertImage()` (Source/x11/convert.c): it builds a known RImage, converts it to a pixmap for the context's visual, reads it back with `XGetImage`, and checks the result — exact on an 8-bit-per-channel TrueColor visual, and readable on lower-depth/palette visuals (which are lossy by design). The wraster sources are compiled in directly, so it needs no gui-linked bundle.

It needs a running X server: it opens `$DISPLAY` and skips cleanly when there is none, so it can run under a headless server such as Xvfb. To exercise it in CI, `make check` would need to run under `xvfb-run`. This is the first backend test that drives the real X rendering path rather than isolated math, so it is a template for testing more of the x11/art image code the same way.

### DIFF
--- a/Tests/x11/GNUmakefile.preamble
+++ b/Tests/x11/GNUmakefile.preamble
@@ -1,0 +1,23 @@
+# Extra build settings for the x11 backend tests.
+#
+# A backend test compiles a backend source file directly, so it can only build
+# for the backend it belongs to.  config.make names the backend being built
+# (BUILD_SERVER); the X11 headers and libraries are added only for x11, and the
+# tests carry a matching source-level guard (via config.h) so they still build
+# -- and skip -- on every other backend.  convert.m also compiles the wraster
+# image sources, which need the Xext, Xrender and Xmu libraries.
+
+# config.h (used by the source-level guard) sits at the top of the (build) tree.
+ADDITIONAL_INCLUDE_DIRS += -I$(GNUSTEP_BUILD_DIR)/../.. -I../..
+
+-include $(GNUSTEP_BUILD_DIR)/../../config.make
+-include ../../config.make
+
+ifeq ($(BUILD_SERVER),x11)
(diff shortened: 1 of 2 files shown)


DTW-Thalion:
@fredkiefer , this is the first backend test that needs a running X server: it drives RConvertImage against a real display and reads the pixmap back. It opens `$DISPLAY` and skips when there is none, so it is safe as-is, but it only runs where a server is present.

To run these in CI, the test step could go under a virtual framebuffer, either `xvfb-run -a make check` or starting Xvfb and exporting `DISPLAY` before the existing `make check`. Xvfb is widely packaged and the tests skip cleanly without it, so it is low-risk. I kept the test display-agnostic so that how the server is provided stays on the CI side, where you have the runner wired up.

DTW-Thalion:
CI has now run and the two x11 jobs fail at the link step, before any test runs:

```
/usr/bin/ld: cannot find -lXmu
```

The x11 jobs install `libxt-dev` but not libXmu. The test compiles the wraster sources in, and one of them, `context.c`, calls `XmuLookupStandardColormap`, so the test executable needs libXmu at link time. The backend library links lazily and builds without it, which is why the build step stays green.

So building these tests needs `libxmu-dev` on the x11 jobs; `libxext-dev` too if you want the XSHM path compiled in (it is `#ifdef`-ed out otherwise). Running rather than skipping them also needs a display, e.g. `make check` under `xvfb-run` as in the note above. Both are CI-side -- the test itself uses no Xmu directly and skips without a display.

DTW-Thalion:
The x11 CI jobs are red here because they lack libXmu and an X server, not because of this change. #95 provides both (libXmu/libXext and xvfb-run); once it lands this needs a rebase onto master to go green.

fredkiefer (approved):

PR #99, opened 2026-07-11
tests: cover the art software compositing operators

DTW-Thalion:
Adds `Tests/art/compositing.m`, covering the software compositing kernels in `Source/art/blit.m`. blit.m has no template of its own -- `blit-main.m` defines pixel-format macros and includes it once per format -- so the test does the same for the 24-bit RGB format and drives the Porter-Duff operators (source-over, source-in, source-out) directly on small pixel buffers. The fully opaque and transparent cases are exact; the partial-coverage cases are checked against the operators' own fixed-point scaling. The kernels are plain integer arithmetic with no libart, X or backend dependency, so the test builds and runs on every configuration.

Writing the test turned up one inconsistency: every colour-scaling expression in blit.m rounds with `+ 0xff` before the `>> 8`, except `sout_oa`, which used `+ 0x80`. The first commit brings it into line with the rest of the file.

### DIFF
--- a/Tests/art/GNUmakefile.preamble
+++ b/Tests/art/GNUmakefile.preamble
@@ -1,0 +1,4 @@
+ADDITIONAL_INCLUDE_DIRS += -I$(GNUSTEP_BUILD_DIR)/../../Source \
+                           -I$(GNUSTEP_BUILD_DIR)/../../Source/art \
+                           -I../../Source -I../../Source/art \
+                           -I$(GNUSTEP_BUILD_DIR)/../.. -I../..
--- a/Tests/art/compositing.m
+++ b/Tests/art/compositing.m
@@ -1,0 +1,165 @@
+/* Tests for the art backend's software compositing kernels in
+ * Source/art/blit.m.
+ *
+ * blit.m has no template of its own; blit-main.m defines a set of pixel-format
+ * macros and includes it once per format.  These tests do the same for the
+ * 24-bit RGB format (separate alpha planes) and then drive the Porter-Duff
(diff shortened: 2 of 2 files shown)


Source/art/blit.m
```diff
@@ -843,9 +843,9 @@ static void MPRE(sout_oa) (composite_run_t *c, int num)
 
       BLEND_READ(s, sr, sg, sb)
 
-      dr = (sr * da + 0x80) >> 8;
-      dg = (sg * da + 0x80) >> 8;
-      db = (sb * da + 0x80) >> 8;
+      dr = (sr * da + 0xff) >> 8;
```
fredkiefer: Are you sure this difference was not here on purpose? If I understand it correctly this only influences the rounding. This was half up and now it is all up. Either way is fine for me.

fredkiefer:
I understand that we could run the art test for every backend, but I would prefer if we really only tested code that is specific for the actual backend we are building. This in not true for the current state and the downside is that the art code would not be tested by the CI, but on the plus side we also would only get fails when the current backend is affected.

Source/art/blit.m
DTW-Thalion: Looked at it again, I think you are right - it was intentional - reverted.

DTW-Thalion:
Right, the 0x80 was deliberate. sout_oa rounds to nearest like the opaque-alpha operators around it (the 0xff ones are the earlier over/in group), so I've dropped the blit.m change altogether. The PR is now just the test, built only when the art backend is the one being built, and the sout_oa check expects the 0x80 rounding.

fredkiefer (approved):

PR #103, opened 2026-07-11
tests: cover the GSStreamContext PostScript operators

DTW-Thalion:
Tests for the GSStreamContext PostScript backend.

Tests/gsc/streamops.m drives one operator from each GSStreamContext family -- colour, line-attribute, matrix and CTM, path and clip, paint, text-show, graphics-state and composite, plus charpath -- to a PostScript file and checks the emitted lines: the right operator name, the arguments in the right order, and numbers formatted locale-independently.

The test needs a backend loaded, so it is built for the cairo backend and skips when there is no display.


DTW-Thalion:
@fredkiefer doing some PR hygiene on this - removing some tests and putting them into a seperate PR.

fredkiefer (approved):

PR #106, opened 2026-07-11
tests: cover more gscolors colour-space conversions

DTW-Thalion:
More coverage for the device colour-space conversions in Source/gsc/gscolors.c.

The existing test covers gray, rgb and the rgb round trips.  This adds the paths it misses: hsb to rgb across all six hue sextants (the earlier test only hits the sextant boundaries), the conversions that reach gray, cmyk and hsb through the rgb fall-through paths, and a cmyk colour with only black.

### DIFF
--- a/Tests/gsc/gscolors.m
+++ b/Tests/gsc/gscolors.m
@@ -134,6 +134,61 @@ main(void)
     PASS(round == YES, "rgb -> cmyk -> rgb round-trips");
   }
 
+  /* --- hsb -> rgb across all six hue sextants (saturation and value 1) --- */
+  {
+    float expect[6][3] = {
+      {1.0, 0.5, 0.0}, {0.5, 1.0, 0.0}, {0.0, 1.0, 0.5},
+      {0.0, 0.5, 1.0}, {0.5, 0.0, 1.0}, {1.0, 0.0, 0.5}
+    };
+    int i;
+
+    for (i = 0; i < 6; i++)
+      {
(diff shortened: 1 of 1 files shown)


Tests/gsc/gscolors.m
```diff
@@ -134,6 +134,63 @@
...
+      {0.0, 0.5, 1.0}, {0.5, 0.0, 1.0}, {1.0, 0.0, 0.5}
+    };
+    int i;
+    BOOL ok = YES;
+
+    for (i = 0; i < 6; i++)
+      {
+	gsMakeColor(&c, hsb_colorspace, (2 * i + 1) / 12.0, 1.0, 1.0, 0);
+	gsColorToRGB(&c);
+	if (!isRGB(c, expect[i][0], expect[i][1], expect[i][2]))
```
fredkiefer: This line looks a bit strange. Why not have the condition in the PASS call itself, like in all the other cases? And if you really want it separated, why not assign the value directly to `ok`?
```
ok = isRGB(c, expect[i][0], expect[i][1], expect[i][2]);
```

It isn't a real issue, it just stands out from the rest of the code.

DTW-Thalion: Yes, you're right. Will fix.

DTW-Thalion: Done, moved the check into the PASS call per sextant.

fredkiefer (approved):

PR #108, opened 2026-07-11
tests: cover the GSGState CTM and coordinate mapping

DTW-Thalion:
Tests for the CTM and coordinate handling in the base GSGState in Source/gsc/GSGState.m.

Tests/gsc/gstate.m drives the matrix operators -- translate, scale, rotate, concat, initmatrix, GSSetCTM and GSConcatCTM -- and checks that a point or a delta is mapped through the current transform as expected, including that the operators compose in the right order.

It needs a backend loaded, so it is built for the cairo backend and skips when there is no display.

### DIFF
--- a/Tests/gsc/GNUmakefile.preamble
+++ b/Tests/gsc/GNUmakefile.preamble
@@ -19,3 +19,5 @@ streamimage_TOOL_LIBS += -lgnustep-gui
 streambezier_TOOL_LIBS += -lgnustep-gui
 # streamops does the same.
 streamops_TOOL_LIBS += -lgnustep-gui
+# gstate does the same.
+gstate_TOOL_LIBS += -lgnustep-gui
--- a/Tests/gsc/gstate.m
+++ b/Tests/gsc/gstate.m
@@ -1,0 +1,147 @@
+/* Tests for the CTM and coordinate handling in the base GSGState in
+ * Source/gsc/GSGState.m: the matrix operators build the current transform, and
+ * a point or delta is mapped through it.
+ *
+ * GSGState lives in the backend bundle, so the test needs a backend loaded
+ * (hence a window server); it opens the display named by the environment and
+ * skips when there is none.  The code under test is the same for every backend;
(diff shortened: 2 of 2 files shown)


Tests/gsc/GNUmakefile.preamble
```diff
@@ -19,3 +19,5 @@ streamimage_TOOL_LIBS += -lgnustep-gui
 streambezier_TOOL_LIBS += -lgnustep-gui
 # streamops does the same.
 streamops_TOOL_LIBS += -lgnustep-gui
+# gstate does the same.
+gstate_TOOL_LIBS += -lgnustep-gui
```
fredkiefer, line 23: Maybe instead of adding this line for each test case, we could try to link in gui with a general statement? I will have to check in GNUstep make if there are options for that. And this could also be changed later, so it should not block merging.

Tests/gsc/gstate.m
```diff
@@ -0,0 +1,147 @@
...
+
+  if (getenv("DISPLAY") == NULL || *getenv("DISPLAY") == '\0')
+    {
+      NSLog(@"no window server available; skipping GSGState CTM tests");
+      DESTROY(pool);
+      return 0;
+    }
+
+  [NSApplication sharedApplication];
+  ctxt = [[NSClassFromString(@"GSStreamContext") alloc] initWithContextInfo:
```
fredkiefer, line 69: Why do we have to use these indirection via `NSClassFromString`? Is it possible to include the headers of these classes here? That way we also wouldn't need to declare the methods in an interface.

fredkiefer (approved):

Tests/gsc/GNUmakefile.preamble
rfm: ADDITIONAL _TOOL_LIBS

fredkiefer: Thank you! That really would simplify things and resolve all the conflicts in that file.

PR #112, opened 2026-07-11
tests: cover the cairo backend compositing operators

DTW-Thalion:
Tests for the cairo backend's compositing operators.

Tests/cairo/compositeops.m fills a red destination, fills again with a given NSCompositingOperation, and reads the pixel back: copy replaces the destination, source-over shows an opaque source, destination-over keeps the opaque destination, plus-lighter adds the source, and clear erases the destination.  This exercises the mapping from NSCompositingOperation to the cairo operator in the cairo graphics state.

It needs a window server (to load the backend), so it opens the display named by the environment and skips when there is none, and guards on the cairo graphics backend being built.

### DIFF
--- a/Tests/cairo/compositeops.m
+++ b/Tests/cairo/compositeops.m
@@ -1,0 +1,111 @@
+/* Tests the cairo backend's compositing operators.  A red destination is
+ * filled, then a second fill is made with a given NSCompositingOperation, and
+ * the resulting pixel is read back with -initWithFocusedViewRect: and checked.
+ * This exercises the operator mapping (NSCompositingOperation to the cairo
+ * operator) in the cairo graphics state.
+ *
+ * It needs a window server (to load the backend), so it opens the display named
+ * by the environment and skips when there is none, and it guards on the cairo
+ * graphics backend being the one built.
+ */
+#import <Foundation/NSObject.h>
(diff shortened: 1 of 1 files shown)


Tests/cairo/compositeops.m
```diff
@@ -0,0 +1,113 @@
...
+  return [rep autorelease];
+}
+
+static BOOL
+pixelIs(NSBitmapImageRep *rep, int x, int y, int r, int g, int b)
+{
+  unsigned char *d = [rep bitmapData];
+  long bpr = [rep bytesPerRow];
+  long spp = [rep samplesPerPixel];
+  unsigned char *px = d + y * bpr + x * spp;
```
fredkiefer: This code makes a lot of assumptions about the structure of the NSBitmapImageRep. Why don't we use the method -getPixel::: to get these values?

DTW-Thalion: Yes, that makes more sense - I was down in the weeds sorting out this test and should have cleaned this up - thanks !

DTW-Thalion: Done, switched to -getPixel:atX:y:.

fredkiefer (approved):

PR #113, opened 2026-07-11
cairo: regression test for translucent image compositing

DTW-Thalion:
The 32-bit image path in the cairo backend rearranged the RGBA bytes into a CAIRO_FORMAT_ARGB32 buffer but left the components straight-alpha, while that format is premultiplied by alpha.  A translucent image therefore came out too bright: a 50% gray drawn over black gave 128 instead of 64.

Read the components per byte, premultiply them, and pack the ARGB32 word.  Reading per byte also makes the packing the same on both byte orders, so the endian-specific branch is no longer needed.  Tests/cairo/imagealpha.m draws a half-transparent gray over black and over white and checks the composited pixel, with an opaque image as a regression guard.

### DIFF
--- a/Tests/cairo/imagealpha.m
+++ b/Tests/cairo/imagealpha.m
@@ -1,0 +1,121 @@
+/* Tests that the cairo backend draws a semi-transparent image correctly.  A
+ * straight-alpha bitmap has to be premultiplied before it is handed to cairo,
+ * whose ARGB32 format is premultiplied; without that a translucent image comes
+ * out too bright.  The test draws a 50% gray over a known background and checks
+ * the composited pixel.
+ *
+ * It needs a window server (to load the backend), so it opens the display named
+ * by the environment and skips when there is none, and it guards on the cairo
+ * graphics backend being the one built.
+ */
+#import <Foundation/NSObject.h>
+#import "Testing.h"
+#include "config.h"
+
+#if defined(BUILD_GRAPHICS) && defined(GRAPHICS_cairo) \
+  && BUILD_GRAPHICS == GRAPHICS_cairo
+
+#import <AppKit/AppKit.h>
+#include <stdlib.h>
+
+static BOOL
+pixelIs(NSBitmapImageRep *rep, int x, int y, int r, int g, int b)
+{
+  NSUInteger px[5];
+
+  [rep getPixel: px atX: x y: y];
+  return (abs((int)px[0] - r) <= 2
+          && abs((int)px[1] - g) <= 2
+          && abs((int)px[2] - b) <= 2);
+}
+
+/* An n x n bitmap filled with one RGBA colour. */
+static NSImage *
+solidImage(int n, int r, int g, int b, int a)
+{
+  NSBitmapImageRep *rep = [[NSBitmapImageRep alloc]
+    initWithBitmapDataPlanes: NULL pixelsWide: n pixelsHigh: n
+                bitsPerSample: 8 samplesPerPixel: 4 hasAlpha: YES isPlanar: NO
+               colorSpaceName: NSDeviceRGBColorSpace
+                 bitmapFormat: NSAlphaNonpremultipliedBitmapFormat
+                  bytesPerRow: n * 4 bitsPerPixel: 32];
+  unsigned char *d = [rep bitmapData];
+  NSImage *img;
+  int i;
+
+  for (i = 0; i < n * n; i++)
+    {
+      d[i * 4 + 0] = r;
+      d[i * 4 + 1] = g;
+      d[i * 4 + 2] = b;
+      d[i * 4 + 3] = a;
+    }
+  img = [[NSImage alloc] initWithSize: NSMakeSize(n, n)];
+  [img addRepresentation: rep];
+  [rep release];
+  return AUTORELEASE(img);
+}
+
+/* Draw IMG over a solid background of the given gray and read the centre. */
+static NSBitmapImageRep *
+over(NSImage *img, int bg)
+{
+  int w = 20, h = 20;
+  NSImage *dst = [[NSImage alloc] initWithSize: NSMakeSize(w, h)];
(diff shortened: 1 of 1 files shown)


fredkiefer:
I am not really sure about this change. There are two aspects I don't like. One is that this assumes that the passed in data is not premultiplied where in some cases it will already be in that format (see the -draw method on NSBitmapImageRep in gui) and then we end up with the wrong values again. What we need here is the bitmapformat, which is not passed to this method. In many cases this will be 0 as the bitmap was converted to one with a format of 0 in -GSDrawImage:: , but as the format is not checked in -isCompatibleBitmap:. This is not your mess, it is mine, but we have to be careful to clean it up and not make things worse.

The other issue I am having is that I am not sure whether the endianness handling is correct in the new code. But then I was never sure whether the old code was correct either :-(
The only thing helping here are real checks by having Riccardo run this code on some of his more unusual machines.

rfm:
My understanding is that RGBA data is to be taken as a sequence of bytes rather than 32bit words, and byte order should therefore be independent of processor architecture.    But I agree that the only way to be sure is to test against reference implementations.
  

  

  

  
  
  
  
  
>

DTW-Thalion:
Exactly - and so I need a baseline behavior - if MacOS is the correct correctness behavior, then a can regress the test in a MacOS runner and make sure we get to the right spot.  @rmottola can you help?

fredkiefer:
I was wrong about  "the format is not checked in -isCompatibleBitmap:". As your other PR proves (and looking at the code would have shown), we make sure that the format is 0.  So on the main code path we should be sure about possible formats.

Tests/cairo/imagealpha.m
```diff
@@ -0,0 +1,120 @@
...
+          && abs((int)px[1] - g) <= 2
+          && abs((int)px[2] - b) <= 2);
+}
+
+/* An n x n bitmap filled with one RGBA colour. */
+static NSImage *
+solidImage(int n, int r, int g, int b, int a)
+{
+  NSBitmapImageRep *rep = [[NSBitmapImageRep alloc]
+    initWithBitmapDataPlanes: NULL pixelsWide: n pixelsHigh: n
```
fredkiefer, line 37: Now I see where the problem with the premultiply comes from. You are using the old method to create an NSBitmapImageRep. This does not include the bitmapFormat, so we treat this as being 0. Which means premultiplied, but the values you pass in aren't. This explains why the results you are getting are to bright.
You should supply the bitmapFormat here and check again.

DTW-Thalion: OK, will look at that again shortly.

DTW-Thalion: Confirmed, thanks. The rep was created with format 0, so the straight values were treated as premultiplied.

Supplying `NSAlphaNonpremultipliedBitmapFormat` then makes it too dark (32 instead of 64 for a 50% grey over black), because `-[NSBitmapImageRep draw]` already calls `-_premultiply` before `GSDrawImage`, and that premultiplies non-premultiplied reps. The premultiply I added to `DPSimage` is a second one.

Centre pixel for a 50% grey over black (should be 64): format 0 with the `DPSimage` premultiply gives 64 (only because `_premultiply` skips it), non-premultiplied with it gives 32, format 0 without it gives 128, non-premultiplied without it gives 64.

So the premultiply already happens in `-_premultiply` and the backend change is redundant; the "too bright" was the test mislabelling the rep. I'll drop the `CairoGState` change and keep the test with the corrected format, which gives 64/191 over black/white - unless you disagree?

fredkiefer: Sounds like the way to go.

DTW-Thalion:
Done. Dropped the CairoGState change; this is now just the test, with the rep created non-premultiplied. It composites to 64 over black and 191 over white.

rmottola:
@DTW-Thalion is there something left to test? I got cited, but tried compiling today with access to big endian systems.
I see that only the test is left. This is on NetBSD/sparc64:

```
--- Running tests in art ---
--- Running tests in cairo ---

cairo/shmbuffer.m:
Skipped set:       shmbuffer.m 75 ... back is not built with the wayland+cairo backend
--- Running tests in gsc ---
--- Running tests in x11 ---

x11/convert.m:
Failed file:  convert.m aborted without running all tests!
--- Running tests in xlib ---

 258 Passed tests
   1 Skipped set
   1 Failed file
```

I see only one failed test and it is not yours. I wanted to compare on classic intel linux, but back runs no test apparently, I will investigate.

DTW-Thalion:
> @DTW-Thalion is there something left to test? I got cited, but tried compiling today with access to big endian systems. I see that only the test is left. This is on NetBSD/sparc64:
> 
> ```
> --- Running tests in art ---
> --- Running tests in cairo ---
> 
> cairo/shmbuffer.m:
> Skipped set:       shmbuffer.m 75 ... back is not built with the wayland+cairo backend
> --- Running tests in gsc ---
> --- Running tests in x11 ---
> 
> x11/convert.m:
> Failed file:  convert.m aborted without running all tests!
> --- Running tests in xlib ---
> 
>  258 Passed tests
>    1 Skipped set
>    1 Failed file
> ```
> 
> I see only one failed test and it is not yours. I wanted to compare on classic intel linux, but back runs no test apparently, I will investigate.

I think you are off the hook - Fred and I sorted it out, as long as the test doesn't show any error for you from my change I think we're OK.

fredkiefer (approved):

rmottola:
I didn't test actually the correct merge. I see a lot of new tests too, for safety, I rerun today:

```
 300 Passed tests
  18 Skipped sets
   1 Failed file

```

All cairo skipped tests are either wayland or win32, so I suppose no issues detected for cairo on the test suite.

Only failure is and was before:
```
x11/convert.m:
Failed file:  convert.m aborted without running all tests!

```

so no new regressions. 👍🏻

PR #114, opened 2026-07-11
tests: cover cairo image orientation and the 24-bit path

DTW-Thalion:
Tests for how the cairo backend places image pixels and decodes pixel formats.

Tests/cairo/imagepixels.m draws a four-colour image and checks it keeps its orientation, with the top of the bitmap at the top, and draws a 24-bit rgb image without alpha and checks it decodes to its colour.  Nearest-neighbour sampling keeps each source pixel a solid block for the read-back.

It needs a window server (to load the backend), so it opens the display named by the environment and skips when there is none, and guards on the cairo graphics backend being built.

### DIFF
--- a/Tests/cairo/imagepixels.m
+++ b/Tests/cairo/imagepixels.m
@@ -1,0 +1,127 @@
+/* Tests that the cairo backend draws image pixels in the right place and
+ * decodes the pixel formats: a four-colour image keeps its orientation (the top
+ * of the bitmap stays at the top), and a 24-bit RGB image without alpha decodes
+ * to its colour.
+ *
+ * It needs a window server (to load the backend), so it opens the display named
+ * by the environment and skips when there is none, and it guards on the cairo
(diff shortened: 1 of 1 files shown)


Tests/cairo/imagepixels.m
```diff
@@ -0,0 +1,127 @@
...
+      DESTROY(pool);
+      return 0;
+    }
+
+  [NSApplication sharedApplication];
+
+  /* A 2x2 image, bitmap row 0 (top) = red, green; row 1 (bottom) = blue, white.
+   * Drawn over the canvas it must keep that layout. */
+  rep = [[NSBitmapImageRep alloc]
+    initWithBitmapDataPlanes: NULL pixelsWide: 2 pixelsHigh: 2
```
fredkiefer, line 78: Again the format is implicitly set to premultiplied alpha, but your data doesn't look that way.

fredkiefer (approved):

PR #115, opened 2026-07-11
back/cairo: return the advance from CairoFontInfo's widthOfString

DTW-Thalion:
CairoFontInfo's widthOfString returned the ink bounding box width (the cairo text extents width) rather than the advance, so a string was reported narrower than the space it occupies.  For a thin character like i the reported width was well under a single character, and the width of a run of characters did not add up to a multiple of the single-character width.  A monospace font even reported different widths for i and W.

Return the advance (x_advance), as advancementForGlyph already does.  Tests/cairo/fontmetrics.m checks that a run of one character is a whole multiple of the single-character width, that the metrics scale with the point size, and that the ascender and descender have the right signs.


fredkiefer (approved):
Great finding!

PR #126, opened 2026-07-11
ci: build and test the art backend

DTW-Thalion:
The art backend is not built by any CI job, so the art-guarded test code in
Tests/art never runs. This adds an art entry to the Linux matrix, with
libart-2.0-dev and the same x11 runtime packages the xlib job uses, so the
art backend is built and its tests run under xvfb.

On my fork the new job builds the art backend and runs the art compositing
tests (9 assertions) that the other jobs skip; the other three jobs are
unaffected.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,6 +36,8 @@ jobs:
             graphics: cairo
           - server: x11
             graphics: xlib
+          - server: x11
+            graphics: art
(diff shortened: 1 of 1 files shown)


DTW-Thalion:
@fredkiefer if you can prioritize this .yaml change that would be great.

fredkiefer (approved):

PR #129, opened 2026-07-13
tests: cover the wayland backend window operations

DTW-Thalion:
Adds a test for the Wayland display server window operations -window::::,
-windowbounds:, -setwindowlevel:: / -windowlevel:, -styleoffsets::::: and
-handlesWindowDecorations.

A window is created through the GSDisplayServer interface and its attributes are
read back: the window number is positive and unique, a zero-area frame does not
produce a zero-size window, -windowbounds: reports the requested size and x
origin, a set window level is read back, and the undecorated backend reports no
style offsets.

It builds only for the wayland+cairo backend and skips when no compositor is
reachable, so it stays inert on the other backends and in CI.

### DIFF
--- a/Tests/cairo/waylandwindowops.m
+++ b/Tests/cairo/waylandwindowops.m
@@ -1,0 +1,144 @@
+/* Coverage for the Wayland display server's window operations: -window::::,
+ * -windowbounds:, -setwindowlevel::, -windowlevel:, -styleoffsets::::: and
+ * -handlesWindowDecorations on WaylandServer (Source/wayland/WaylandServer.m).
+ *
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):
I don't see why any of these operations is Wayland specific. That should work for the other servers just the same.

DTW-Thalion:
Actually it's a fair point. I was just working in the Wayland backend when I produced them so labelled them that way. Arguably  one thing I could do after all the testing touchpoints are in place, is rationalize the testing because I do know I have been more focused on checking off the boxes of completeness per backend, than whether I can economize on unit tests. I will make a note to come look at it after I get the other things on my list sorted.

PR #130, opened 2026-07-13
back/wayland: correct the y origin returned by -windowbounds:

DTW-Thalion:
The y origin was computed as the screen height minus the stored top-down
position, leaving it one window height too high, so the reported bounds did
not match the frame the window was created with.  Convert the stored position
back to screen coordinates with WaylandToNS(), and add a round-trip test.


DTW-Thalion:
@fredkiefer this PR is fairly high priority because it is required for a number of other PRs for Wayland that I am working on, so if you can look at this one first, that would be great.

fredkiefer (approved):

PR #139, opened 2026-07-13
tests: cover the fontconfig face info

DTW-Thalion:
Adds a test for the fontconfig face descriptor and the character set it exposes.

FCFaceInfo carries a family name, weight and traits through their getters and
setters, has a fixed cache size, and builds a display name from the pattern: the
family alone, or the family and style together.  -characterSet resolves the
pattern to a font and wraps its coverage, reporting ASCII letters as members and
an unassigned code point as not a member.

It builds only for the cairo graphics backend and skips when no window server is
available, as the other font tests do.

### DIFF
--- a/Tests/cairo/faceinfo.m
+++ b/Tests/cairo/faceinfo.m
@@ -1,0 +1,168 @@
+/* Coverage for the fontconfig face descriptor (Source/fontconfig/FCFaceInfo.m)
+ * and the character set it exposes (FontconfigCharacterSet in
+ * Source/fontconfig/FCFontEnumerator.m).
+ *
+ * FCFaceInfo carries a family name, weight and traits (each with a getter and a
+ * setter), a fixed cache size, and a display name built from the pattern: the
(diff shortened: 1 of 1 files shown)


Tests/cairo/faceinfo.m
```diff
@@ -0,0 +1,168 @@
...
+
+  Class faceClass = Nil;
+  id gen = nil;
+
+  NS_DURING
+    {
+      [NSApplication sharedApplication];
+      faceClass = NSClassFromString(@"FCFaceInfo");
+      gen = [[[NSClassFromString(@"FontconfigPatternGenerator") alloc] init]
+	      autorelease];
```
fredkiefer, line 58: You should go back to using the Macros.

fredkiefer (approved):

Tests/cairo/faceinfo.m
DTW-Thalion: Some habits die hard... I'll have to do a pass and fix them.

PR #140, opened 2026-07-13
back/cairo: give cairo a UTF-8 filename for PS and PDF output

DTW-Thalion:
cairo_ps_surface_create and cairo_pdf_surface_create take a UTF-8 char* file
name, but -fileSystemRepresentation returns a UTF-16 unichar* on Windows, so the
PS and PDF surfaces did not compile there.  Use -UTF8String, which is a UTF-8
char* on every platform.


fredkiefer (approved):
I realy wasn't aware that Cairo uses UTF-8 even in Windows.

PR #141, opened 2026-07-13
tests: cover the win32 backend screen queries

DTW-Thalion:
Adds a test for the win32 display server queries -screenList,
-boundsForScreen:, -windowDepthForScreen: and -availableDepthsForScreen:.

These are answered from the desktop monitors without mapping a window, so the
test checks the GSDisplayServer contract: a non-empty list of screen numbers, a
positive-size rect for each screen and NSZeroRect for a screen that is not
present, a valid RGB window depth, and a zero-terminated list of available
depths that includes the screen depth.


fredkiefer (approved):
I don't see how this test differs from the one you wrote for Wayland. I think we should have this for all the servers.

DTW-Thalion:
Agree, I will rationalize them after I have all the coverage confirmed.

PR #142, opened 2026-07-13
tests: cover the win32 backend window style mapping

DTW-Thalion:
Adds a test for -windowStyleForGSStyle: and -exwindowStyleForGSStyle: on the
win32 server, which turn an NSWindow style mask into the Win32 WS_* and WS_EX_*
window styles.

Without server-drawn decorations a window maps to a clipped popup.  With them a
titled window has a caption, a closable one a system menu, a miniaturizable one a
minimize box and a resizable one a sizing border, and every result clips its
children.  The extended style keeps utility and borderless windows out of the
taskbar and drops every window to a tool window frame when the native taskbar is
not used.


fredkiefer (approved):
These are rather implementation specific details. Not sure whether we should test these at all.

PR #143, opened 2026-07-13
back/winlib: fix the dash pattern type in the win32 graphics state

DTW-Thalion:
The dash pattern was declared as an array of CGFloat pointers rather than an
array of CGFloat, so -getLineDash:count:phase: filled it with dash lengths cast
to pointers and the pen pattern was built from the pointer values.  It also did
not compile on a modern compiler.  Declare it as an array of CGFloat.


fredkiefer (approved):
Good catch

PR #144, opened 2026-07-13
tests: cover the winlib backend rendering

DTW-Thalion:
Adds a Tests/winlib directory with a rendering test for the win32 GDI graphics
backend, driven through the AppKit offscreen path: lock focus on an NSImage,
draw, and read the pixels back with -initWithFocusedViewRect:.

It checks that a solid fill paints the image, that filling the left half leaves
the right half alone, and that a filled bezier rectangle paints inside and
leaves the outside alone, which exercises WIN32GState fill and path rendering
and the GSReadRect read-back.

### DIFF
--- a/Tests/winlib/GNUmakefile.preamble
+++ b/Tests/winlib/GNUmakefile.preamble
@@ -1,0 +1,17 @@
+# Extra build settings for the winlib backend tests.
+
+# config.h (used by the source-level guards) sits at the top of the (build)
+# tree.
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
Again a test, where the same code could be used for all backends.
Not sure why you did add the OpenGl library in the Makefile, for this specific test it wouldn't be needed.

DTW-Thalion:
@fredkiefer will do PR hygiene on all of these - check out my Issue #167 for my questions to you on this.

PR #145, opened 2026-07-13
tests: cover the winlib backend compositing

DTW-Thalion:
Adds a compositing test for the win32 GDI graphics backend, driven through the
AppKit offscreen path.

A Copy fill replaces the destination, a source-over fill with a half-transparent
colour blends towards it, and compositing an opaque image with Copy or
source-over paints the image, which exercises WIN32GState fill compositing and
its image compositing through the BitBlt and AlphaBlend routes with the
GSReadRect read-back.


fredkiefer (approved):
There is nothing Windows specific in this test. We should rather have it in libs-gui as a general test for all backends.

DTW-Thalion:
Noted.

PR #146, opened 2026-07-13
tests: cover the win32 backend window operations

DTW-Thalion:
Adds a test for the win32 display server window operations -window::::,
-windowbounds:, -setwindowlevel:: / -windowlevel: and -styleoffsets:::::.

A hidden window is created through the GSDisplayServer interface and its
attributes are read back: the window number is positive and unique,
-windowbounds: reports the frame the window was created with, a set window
level is read back, and -styleoffsets: reports the decoration insets of a
titled window with a taller top for the title bar, or no insets when the
server does not draw decorations.


fredkiefer (approved):
Again not much Windows specific.

PR #147, opened 2026-07-13
tests: cover the winlib backend image drawing

DTW-Thalion:
Adds an image drawing test for the win32 GDI graphics backend.

A four-colour image keeps its orientation when drawn, with the top of the
bitmap staying at the top, and a 24-bit RGB image without alpha decodes to its
colour, which exercises WIN32GState -DPSimage: through GSCreateBitmap and the
PlgBlt.


fredkiefer (approved):
Again, nothing Windows specific, we should have this test in gui

PR #148, opened 2026-07-13
tests: cover the winlib backend stroke and dashes

DTW-Thalion:
Adds a stroke test for the win32 GDI graphics backend.

A solid stroked line paints black along its centre and leaves the pixels clear
of it white, and a dashed stroke paints its first dash and leaves the following
gap white, which exercises WIN32GState stroke path and the ExtCreatePen pen
setup including the dash pattern.


fredkiefer (approved):
Again, this test belongs into gui

DTW-Thalion:
Added to the list.

PR #149, opened 2026-07-13
tests: cover the winlib backend line caps and joins

DTW-Thalion:
Adds a line cap and join test for the win32 GDI graphics backend.

A butt cap ends at the endpoint while a square cap extends past it by half the
line width, and a miter join fills the sharp outer corner more than a bevel
join, which exercises WIN32GState pen end-cap and join setup.


fredkiefer (approved):
Same as the last few tests, does not belong here

DTW-Thalion:
Added to the list.

PR #150, opened 2026-07-13
tests: cover the win32 font enumeration

DTW-Thalion:
Adds a test for the win32 GDI font enumerator, the font discovery layer the
winlib backend uses through GSFontEnumerator.

The checks are font independent: the enumerator lists at least one family and
one font, a listed family has a member, and the default system, bold and
fixed-pitch names are set with the three roles differing.


fredkiefer (approved):
again a gui test

PR #151, opened 2026-07-13
tests: cover the win32 font metrics

DTW-Thalion:
Adds a test for the win32 GDI font metrics, reached through NSFont.

A proportional font advances a wide glyph further than a narrow one, the
advancement scales with the point size, a fixed-pitch font advances every glyph
the same, a glyph has a positive-size bounding box, and a common glyph is
encoded.  A small tolerance allows for GDI rounding advances to whole pixels.


fredkiefer (approved):
Rather a gui test

PR #155, opened 2026-07-13
tests: cover the wayland EGL OpenGL context

DTW-Thalion:
Adds a test for the Wayland EGL/OpenGL support.

An NSOpenGLPixelFormat and NSOpenGLContext are created, and an NSOpenGLView
backs a window with a current context that reports a GL version, which
exercises WaylandGLPixelFormat EGL config selection and WaylandGLContext EGL
context and wl_egl_window surface.

It builds only for the wayland+cairo backend with EGL and skips when the backend
or a GL drawable cannot be reached.  It needs libGL, which the preamble links
for that test.


fredkiefer (approved):

DTW-Thalion:
The `ci (wayland, cairo)` failure this introduced is fixed by #169.

The preamble linked `-lGL` into `wlglcontext` for every wayland+cairo build. The CI container has EGL/GL only as weston runtime dependencies, not the development libraries, so `configure` finds no EGL, the test compiles to its skip stub, and the tool still fails to link against the absent libGL. #169 gates the `-lGL` on `WITH_EGL` so it is only linked where the EGL path is built.

PR #156, opened 2026-07-13
tests: cover the wayland window surface rendering

DTW-Thalion:
Adds a rendering test for the Wayland shared-memory window surface.

Unlike the offscreen rendering test, this draws into a real window content view
backed by the wl_shm surface and reads the pixels back: a solid fill paints the
window and filling the left half leaves the right half alone, which exercises
the shm surface cairo drawing and the GSReadRect read-back through the on-window
path.

It builds only for the wayland+cairo backend and skips when no compositor can be


fredkiefer (approved):
Most of this tests should either be moved to gui or into a shared directory to execute them for all the different backends.

PR #158, opened 2026-07-14
back/cairo: keep the exact clip across a graphics state save

### LINKED ISSUE #68: saveGraphicsState erases clipping path in Cairo
This issue showed up as NSGradient not being draw correctly w.r.t. the clipping path.  See
   https://github.com/gnustep/libs-gui/issues/228

Problem:
  saveGraphicsState/restoreGraphicsState is not handling the clipping path correctly.
  It seems that saveGraphicsState is actually clearing the clipping path

Inheritance Hierarchy (for reference)

```
    CairoContext : GSContext : NSGraphicsContext

    CairoGState : GSGState : NSObject
```

Why is saveGraphicsState clearing the clipping path?  The call-chain is complex, but here is what is happening.

When saveGraphicsState is called, it works its way up the inheritance hierachy

    CairoContext : GSContext : NSGraphicsContext

to NSGraphicsContext, since none of the intermediate classes have implemented it.

In NSGraphicsContext, the default implementation of saveGraphicsState trampolines to method DPSgsave, which is then defined as subclassResponsiblity.

The subclass method that handles it is `[GSContext DPSgsave]` which does this: the current state becomes a deep copy of the graphics context.

```
- (void) DPSgsave
{
  ctxt_push(gstate, gstack);
  AUTORELEASE(gstate);
  gstate = [gstate copy];
}
```

So this is where the graphics state is actually pushed.  And a  "copy" of the current state becomes the current state.  But the copy does not have the clipping path.

GSGState implements "deepen" which does a copyWithZone on a bunch of stuff, including the CairoGraphicsState.

The copyWithZone of CairoGState tries to copy the cairo context by making a new cairo
context and copying over its elements.  However, it can't copy over the clipping path
because there is no clean way to access that from Cairo.  So it is lost.
(Cairo really wants its context to be an opaque data structure, and
trying to copy one piecemeal doesn't seem to work.)

At this point, we have an "almost"-copy of the Cairo state, but not quite.  The clipping
path is lost.

APPROACH to a SOLUTION:

Cairo implements a perfectly good save/restore mechanism on its own.  And rather than
creating new cairo contexts and copying over the internals, we could use save/restore
directly.  NSGraphicsContext requires the same stack ordering of save/restore, so the methods
should map directly.

In file CairoContext.m: add saveGraphicsState and restoreGraphicsState that use
the cairo methods.  By not calling "super" in these methods the entire existing
copy-push/pop generic mechanism is avoided, as far as I can tell, and all of the
above description of the call-chain does not happen.

```
@implementation CairoContext

- (void) saveGraphicsState {
  if (CGSTATE->_ct)
    cairo_save(CGSTATE->_ct);
}

- (void) restoreGraphicsState {
  if (CGSTATE->_ct)
    cairo_restore(CGSTATE->_ct);
}
@end
```

This has been tested with Gworkspace, Terminal, TextEdit and the complex gradients of IconApp and seems to work.  Do not know the assumptions made during the current implementation, however.

Image

DTW-Thalion:
Fixes #68.

`-[CairoGState copyWithZone:]` rebuilt the clip from `cairo_copy_clip_rectangle_list`, which reports a non-rectangular clip as `CLIP_NOT_REPRESENTABLE` and fell back to the clip bounding box. A clip set before `-saveGraphicsState` was widened to its bounding box on the copied state, so an NSGradient drawn in a bezier path spilled outside the shape (libs-gui #228).

Track the clip paths in base space as they are applied in `-DPSclip` and `-DPSeoclip`, drop them in `-DPSinitclip` and `-DPSinitgraphics`, and replay them onto the copied context. This keeps the separate per-state `cairo_t` that `GSDefineGState` and `DPSsetgstate` rely on, so it avoids the scrolling and image regressions the earlier `cairo_save`/`cairo_restore` approach caused.

`Tests/cairo/clipsaverestore.m` covers triangle, oval and even-odd ring clips across save and restore; it fails on master (the non-rectangular clips widen to their bounding box) and passes with the fix. The rest of the `Tests/cairo` suite is unchanged.

### DIFF
--- a/Headers/cairo/CairoGState.h
+++ b/Headers/cairo/CairoGState.h
@@ -36,6 +36,7 @@
   @public
     cairo_t *_ct;
     CairoSurface *_surface;
+    NSMutableArray *_clipPaths;  /* base-space clip paths, in application order */
 }
 
 - (void) GSCurrentSurface: (CairoSurface **)surface : (int *)x : (int *)y;
--- a/Source/cairo/CairoGState.m
+++ b/Source/cairo/CairoGState.m
@@ -113,7 +113,42 @@ static inline cairo_filter_t cairoFilterFromNSImageInterpolation(NSImageInterpol
 }
 
 
-@implementation CairoGState 
+/* Emit a base-space bezier path onto a cairo context, resetting any current
+ * path first.  Used to replay tracked clip paths onto a copied context. */
+static void
+appendBezierToCairo(cairo_t *ct, NSBezierPath *bpath)
+{
+  NSInteger count = [bpath elementCount];
+  NSInteger i;
+
+  cairo_new_path(ct);
+  for (i = 0; i < count; i++)
+    {
+      NSPoint points[3];
+
+      switch ([bpath elementAtIndex: i associatedPoints: points])
+        {
+          case NSMoveToBezierPathElement:
+            cairo_move_to(ct, points[0].x, points[0].y);
+            break;
+          case NSLineToBezierPathElement:
+            cairo_line_to(ct, points[0].x, points[0].y);
+            break;
+          case NSCurveToBezierPathElement:
+            cairo_curve_to(ct, points[0].x, points[0].y,
+                           points[1].x, points[1].y,
+                           points[2].x, points[2].y);
+            break;
+          case NSClosePathBezierPathElement:
+            cairo_close_path(ct);
+            break;
+          default:
+            break;
+        }
+    }
+}
+
+@implementation CairoGState
 
 + (void) initialize
 {
@@ -129,6 +164,7 @@ static inline cairo_filter_t cairoFilterFromNSImageInterpolation(NSImageInterpol
       cairo_destroy(_ct);
     }
   RELEASE(_surface);
+  RELEASE(_clipPaths);
 
   [super dealloc];
 }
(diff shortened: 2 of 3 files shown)


fredkiefer:
This is a very impressive change. I will have to run it locally to see, if there are any downsides.
I think that your test code at the moment won't detect any real changes, just the missing log message. What would be needed is some more drawing after the restore to confirm that the actual clip is still in place and it didn't get extended to a rectangle.
And if it really works we should try to do something similar on the GSGstate (with the NSBezierPath directly) to have that issue fixed for all backends. But I am still a bit reluctant.

DTW-Thalion:
Test away ! I am not exhaustively testing these in live environments, so while they seem to work based on whatever unit tests I can muster and I try functional testing where I can, the devil is in the details, and yes we can look at GSGState if you think this will hold up to scrutiny.

DTW-Thalion:
Just an update: I ran it against both backends to double-check. On the current master backend three of the six fail — the triangle and oval bleed past the real clip into its bounding box after the save, and the even-odd ring drops out; on the branch all six pass. So the fill after the save does land on the copied gstate and exercises the real clip shape, not just the log path.

One gotcha that cost me a while: the loaded backend is the `.cairo` bundle variant, so a stale installed bundle makes old and new render identically and the test looks inert — a clean rebuild/install sorts it.

A draw-after-restore check would land on the restored original gstate, which keeps its clip regardless, so it wouldn't cover `copyWithZone:` the way the save-side fill does.

rmottola:
I will test this trying out real-world usage of applications with your patch applied.

rmottola:
Looks fine, I tested a lot of application with your patch on a cairo backend.

fredkiefer (approved):

PR #165, opened 2026-07-15
Wayland: build pointer enter/leave with enterExitEventWithType:

### LINKED ISSUE #79: Wayland Backend file open dialogs closing prematurely
In Wayland backend when an NSOpenPanel is called,  the open dialog will appear briefly. Moving the cursor over the NSOpenPanel window, or any other window from the parent application causes the NSOpenPanel to immediately close. This does not close when moving the cursor over GNUstep windows from other applications. Behaviour happens on Wayfire and Ambrosia compositors.

DTW-Thalion:
Closes #79 
The Wayland pointer enter and leave handlers build NSMouseEntered and NSMouseExited events with `+[NSEvent mouseEventWithType:]`, which raises `NSInvalidArgumentException` ("mouseEvent with wrong type") because `GSMouseEventMask` does not cover those two types; they belong to `GSEnterExitEventMask`. Both handlers now use `+[NSEvent enterExitEventWithType:...]`, the correct constructor for enter/exit events.

The raise happens inside `wl_display_dispatch` in `-[WaylandServer receivedEvent:]`, a run loop input source. `-[NSApplication run]` wraps event dispatch in NS_DURING/NS_HANDLER, so in normal use the exception is caught and logged and the application continues. `-[NSApplication runModalSession:]` has no such handler and dispatches events in NSModalPanelRunLoopMode, so an enter or leave raised while an NSOpenPanel or NSSavePanel is open unwinds the modal loop and the panel closes. 

This is issue #79: the file dialog closes as soon as the pointer moves over it or over another window of the same application. The same defect also aborts a drag, when the compositor sends `wl_pointer.leave` to take the pointer for the drag grab, which is why this PR is needed before #166.

### DIFF
--- a/Source/wayland/WaylandServer+Cursor.m
+++ b/Source/wayland/WaylandServer+Cursor.m
@@ -87,19 +87,17 @@ pointer_handle_enter(void *data, struct wl_pointer *pointer, uint32_t serial,
       gcontext = GSCurrentContext();
       eventLocation = NSMakePoint(sx, window->height - sy);
 
-      event = [NSEvent mouseEventWithType:NSMouseEntered
+      /* NSMouseEntered must be built with the enter/exit constructor;
+       * +mouseEventWithType: rejects it ("mouseEvent with wrong type"). */
+      event = [NSEvent enterExitEventWithType:NSMouseEntered
 				 location:eventLocation
 			    modifierFlags:wlconfig->modifiers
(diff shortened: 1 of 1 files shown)


DTW-Thalion:
@fredkiefer this PR is needed before #166 can work otherwise it will hang- this one I think can be checked unlike the soaking I would like to see with #166.

fredkiefer (approved):

PR #166, opened 2026-07-15
Wayland: implement drag and drop via wl_data_device

### LINKED ISSUE #72: Drag/Drop icons fails on Wayland Backend
Dragging/dropping icons fails on Wayland Backend. Can reproduce in GWorkspace or GFinder (forked from GWorkspace). When built with Wayland backend if run in Ambrosia or Wayfire Compositors Drag/drop mouse over icons just draws grey outlines and never seems to grab/drag icons.

Image attached, the grey near the SystemPreferences.app icon is what drew when trying to click/drag the icon to the Dock.

Image

DTW-Thalion:
Closes #72.

The Wayland backend had no drag and drop support. `-dragInfo` fell through to the base `GSDragView`, whose client-driven tracking loop moves a drag window to follow the pointer and hit-tests other windows. That cannot work under Wayland, where the compositor owns the pointer during a drag and a client can neither see nor position other surfaces.

This adds a `WaylandDragView` (a `GSDragView` subclass) wired in as the server's `-dragInfo`, backed by `wl_data_device`. The source advertises the drag pasteboard through a `wl_data_source` and calls `wl_data_device_start_drag`, then runs a modal loop until the compositor concludes the drag. The destination turns the compositor's `wl_data_device` enter/motion/drop events into the `GSAppKitDragging` events that `NSWindow` dispatches to the target view, using the drag view as the `NSDraggingInfo`.

Depends on #165; without that fix the drag aborts as soon as the compositor takes the pointer.

Two limitations are left for follow-ups: no drag icon surface is attached yet, so the compositor shows no drag image, and a drop originating in another client is not yet read into an `NSPasteboard`. Same-process drags work.

### DIFF
--- a/Headers/wayland/WaylandServer.h
+++ b/Headers/wayland/WaylandServer.h
@@ -82,6 +82,8 @@ typedef struct _WaylandConfig
   struct wl_shm		*shm;
   struct wl_seat		 *seat;
   struct wl_keyboard	     *keyboard;
+  struct wl_data_device_manager *data_device_manager;
+  struct wl_data_device	     *data_device;
   struct xdg_wm_base	     *wm_base;
   struct zwlr_layer_shell_v1 *layer_shell;
 #ifdef HAVE_EGL
--- a/Source/wayland/GNUmakefile
+++ b/Source/wayland/GNUmakefile
@@ -51,6 +51,7 @@ WaylandServer+Keyboard.m  \
 WaylandServer+Seat.m  \
 WaylandServer+Xdgshell.m  \
 WaylandServer+Layershell.m  \
+WaylandDragView.m  \
 WaylandInputServer.m  \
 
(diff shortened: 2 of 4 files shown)


DTW-Thalion:
@fredkiefer, let this PR soak for a while. It's efffectively a new feature as it fills in functionality that was missing in Wayland to mirror what the x11 backend has to do drag and drop. I am not running on Wayland as a matter of course, so I tested it with a minimal Wayland setup and compositor. So this needs to be stress tested by those better setup than I. This was just an opportunity for me to try building it - so further testing required.

fredkiefer (approved):
I am no expert on Wayland, so I can just hope this works :-)

DTW-Thalion:
If it doesn't I'm sure with some help to debug  I can sort it out.

PR #168, opened 2026-07-21
Render context shadows in the raster backends

DTW-Thalion:
The backends ignored the shadow set through `NSGraphicsContext`, so `[shadow set]` followed by a fill or stroke drew nothing extra. This adds offset-shadow rendering to all four raster backends: cairo, art, xlib and winlib.

The shadow state is stored on the shared `GSGState` (`-setShadow:`/`-shadow`, released and copied with the state), and `GSContext` forwards `-setShadow:`/`-shadow` to it. Because the state lives on the graphics state, saving and restoring the graphics state carries the shadow with it, and no libs-gui change is needed (the gui already calls `[ctxt setShadow:]`).

Each backend paints the shadow before the real path: the current path, offset by the shadow offset, in the shadow colour. This covers the fill, even-odd fill and stroke operators. cairo draws the offset path directly on its context; the other backends render an offset copy of the path in the shadow colour, restoring the fill/stroke colour afterwards. Blur is not applied.

`Tests/cairo/dropshadow.m` draws a shape with a coloured shadow at a known offset through the AppKit offscreen path and checks the offset region for the shadow colour, that the shape sits on top of its own shadow, and that restoring the graphics state clears the shadow. It runs on cairo and art (8 assertions each), whose offscreen drawing reads back reliably in a headless run. The xlib and winlib backends render shadows the same way; xlib was verified locally against a live X server (the headless xlib offscreen path is not reliable enough to assert on in CI) and winlib was not built here.

Addresses gnustep/libs-gui#217

### DIFF
--- a/Headers/gsc/GSGState.h
+++ b/Headers/gsc/GSGState.h
@@ -38,6 +38,7 @@
 @class NSBezierPath;
 @class NSFont;
 @class NSColorSpace;
+@class NSShadow;
 @class GSContext;
 
 typedef enum {
@@ -72,6 +73,7 @@ typedef enum {
   BOOL    _antialias;
   NSPoint _patternPhase;
   NSCompositingOperation _compositingOperation;
+  NSShadow *_shadow;            /* Current shadow, or nil */
 }
 
 - initWithDrawContext: (GSContext *)context;
@@ -89,6 +91,8 @@ typedef enum {
 - (void) setPatternPhase: (NSPoint)phase;
 - (NSCompositingOperation) compositingOperation;
 - (void) setCompositingOperation:(NSCompositingOperation) operation;
+- (void) setShadow: (NSShadow *)shadow;
+- (NSShadow *) shadow;
 
 - (void) compositeGState: (GSGState *)source
                 fromRect: (NSRect)aRect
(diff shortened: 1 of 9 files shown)


DTW-Thalion:
Confirmed this covers the Wayland server. The shadow is rendered in the cairo
graphics backend (CairoGState -_drawShadowForOperation:, which only touches the
cairo context), and the Wayland backend has no graphics gstate of its own, so a
wayland+cairo build draws shadows through the same code as x11+cairo.

Verified with the wayland+cairo backend loaded (GSCurrentServer is
WaylandServer) against a live compositor: dropshadow.m runs and all 8 checks
pass.

The last commit broadens the test's guard to accept WAYLAND_DISPLAY, since it
previously skipped whenever DISPLAY was unset and so did not run in a
Wayland-only session.

fredkiefer:
This code looks like it would be sufficient to have a -_paintPath: method in all the backend and then the actual shadow drawing could be implemented in GSGState. Of course this simplification could also be done later on.

DTW-Thalion:
> This code looks like it would be sufficient to have a -_paintPath: method in all the backend and then the actual shadow drawing could be implemented in GSGState. Of course this simplification could also be done later on.

Leave it with me :)

fredkiefer (approved):

Issue #170, opened 2026-07-22
`default{Bold,System,FixedPitch}FontName` should not hardcode font names


probonopd:
The fontconfig backends (`FCFontEnumerator` and `FTFontEnumerator`) maintain hardcoded priority lists in `defaultSystemFontName`, `defaultBoldSystemFontName`, and `defaultFixedPitchFontName`. For example, from `FCFontEnumerator.m`:

 ```objc
 - (NSString *) defaultBoldSystemFontName
 {
   if ([allFontNames containsObject: @"DejaVuSans-Bold"])
     return @"DejaVuSans-Bold";
   if ([allFontNames containsObject: @"FreeSans-Bold"])
     return @"FreeSans-Bold";
   // ... Tahoma, Arial, etc.
   return @"Helvetica-Bold";
 }
```

This has several problems:

 1. Breaks silently on systems without those fonts.  A user installs a custom font (e.g., Inter) and sets NSFont = "Inter" in user defaults. The regular system font resolves correctly. But NSBoldFont has no user default, so boldSystemFontOfSize: falls through to
 defaultBoldSystemFontName, which returns "Helvetica-Bold" (Inter isn't in the list). Fontconfig can't find "Helvetica-Bold" → bold lookup fails → getNSFont() fallback chain degrades to the regular-weight system font. The application renders bold UI elements in regular weight with no error or warning.

 2. Embeds vendor/foundry preferences into the framework.  DejaVu, Bitstream Vera, FreeSans, Tahoma, Arial, Helvetica — arbitrary choices from 20 years ago. Every distro or desktop that prefers a different default font requires a source patch.

 3. Duplicated across every backend.  FCFontEnumerator (Cairo/Xft), FTFontEnumerator (Art), WIN32FontEnumerator — each has its own copy. They diverge independently.

 4. No extension path.  Adding support for a new font family means editing C source in every backend instead of, say, a plist-based ranking or deriving the bold name from the configured family.

Suggested approach

Instead of hardcoding font names, these methods should derive the bold/fixed-pitch name from the user-configured system font family:

 - defaultSystemFontName → already handled by fontNameForRole via NSFont user default; this method only needs a safe last resort.
 - defaultBoldSystemFontName → take the system font's family name (from the active NSFont default) and append "-Bold". Fall back to a hardcoded list only if the constructed name isn't found in fontconfig.
 - defaultFixedPitchFontName → same approach from NSUserFixedPitchFont.

[...]

PR #172, opened 2026-07-26
back/fontconfig: resolve default bold and fixed-pitch fonts through fontconfig

DTW-Thalion:
Requires: [libs-gui/PR#816](https://github.com/gnustep/libs-gui/pull/816)

`defaultBoldSystemFontName` and `defaultFixedPitchFontName` in the fontconfig enumerator kept hardcoded priority lists of font names. When a user configures a custom system font (the `NSFont` default) without a matching `NSBoldFont`, the bold lookup falls through to an unrelated hardcoded family, fontconfig can't find it, and bold UI text silently renders in regular weight (issue #170, case 1).

Instead of string lists, this asks fontconfig — which is already the enumeration engine here — for the right face:

- **bold**: derive the bold face of the configured system font family via `FC_WEIGHT_BOLD`.
- **fixed-pitch**: match the `monospace` generic.

The match is converted through the existing `faFromFc`, so the returned name is exactly the one under which the font was enumerated (guaranteed present in `allFontNames`). The historic lists remain only as a last resort when fontconfig resolves nothing, so behaviour can only improve — the default (DejaVu) case is unchanged.

Verified on the shared FCFontEnumerator across the graphics backends that use it:
- **cairo (x11)** and **xlib**: with `NSFont` unset, bold/fixed resolve to `DejaVuSans-Bold`/`DejaVuSansMono` as before; with `NSFont` set to a non-default family (Bitstream Charter), bold now correctly derives `CharterBT-Bold` (same family, bold weight) rather than the old `DejaVuSans-Bold`.
- **wayland+cairo**: builds and uses the same graphics-layer resolution (server-independent).

Scope: this fixes the shared fontconfig enumerator (cairo, xlib). The art backend's `FTFontEnumerator` and the Windows `WIN32FontEnumerator` keep their own copies of the hardcoded logic; the same idea applies there (fontconfig for FT, GDI `LOGFONT` weight/pitch for Win32) and is left as a follow-up. The broader architectural cleanup in #170 (de-duplication, plist ranking) is also out of scope here.

Addresses #170.

### DIFF
--- a/Source/fontconfig/FCFontEnumerator.m
+++ b/Source/fontconfig/FCFontEnumerator.m
@@ -453,8 +453,67 @@ static NSArray *faFromFc(FcPattern *pat)
   return @"Helvetica";
 }
 
+/* Return the enumerated font name of fontconfig's best match for pat,
+   applying the usual configuration and default substitutions, or nil.
+   Reuses faFromFc so the returned name is the same one under which the
+   font was enumerated (and thus present in allFontNames).  The caller
+   owns pat. */
+static NSString *
+fcDefaultFontName(FcPattern *pat)
+{
+  NSString *name = nil;
+  FcPattern *match;
+  FcResult result;
+
+  FcConfigSubstitute(NULL, pat, FcMatchPattern);
+  FcDefaultSubstitute(pat);
+  match = FcFontMatch(NULL, pat, &result);
+  if (match != NULL)
+    {
+      NSArray *fontArray = faFromFc(match);
+
+      if (fontArray != nil)
+        {
+          name = [fontArray objectAtIndex: 0];
+        }
+      FcPatternDestroy(match);
+    }
+  return name;
+}
+
 - (NSString *) defaultBoldSystemFontName
 {
+  /* Ask fontconfig for the bold face of the configured system font family,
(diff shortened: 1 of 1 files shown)


DTW-Thalion:
The cairo CI failure (glyphmetrics.m and pdfps.m aborting) is a pre-existing latent bug in libs-gui that this change surfaces: init_font_roles kept the backend's default font names without retaining them, relying on backends returning string constants. Once this enumerator returns fontconfig-derived (non-constant) names, those font_roles[].defaultFont entries dangle after the autorelease pool drains and the next font lookup crashes.

Fixed in gnustep/libs-gui#816 (retain the names with ASSIGN). With #816 installed, the cairo tests here go from intermittent aborts to 0/30 crashes. This PR needs #816 to merge first; the CI clones libs-gui master, so it will go green once #816 lands.

fredkiefer (approved):
Great idea! The code looks like it could be simplified even more, but I will leave that to later. After all your changes are in, I should go through the whole code anyway.

DTW-Thalion:
Took a pass at the simplification you mentioned in #178, where the default font methods now share a helper for their candidate lists.

Issue #45, opened 2023-12-06
Multiple Copy&Paste issues (X11)


rmottola:
Copy&Paste has multiple issues, especially when working with other applications (e.g. GIMP, LibreOffice...) Let me track cumulatively what works and what not.

Works:
- Copy&Paste within PRICE of small/medium sized images
- Copy&Paste within PRICE of big images (just fixed) in a clean status
- Copy from PRICE to GIMP of medium sized image
- Copy from GIMP to PRICE of medium sized image
- Copy from Ink to LibreOffice of Rich Text in a "clean" environment

Does Not work:
- [x] Copy from PRICE to GIMP of big images (after a long while: There is no image data in the clipboard to paste)
- [x] Copy from GIMP to PRICE of big images
- [ ] Copy from PRICE to GIMP of 16bit
- [ ] Copy from GIMP to PRICE of 16bit
- [ ] Copy from LibreOffice to Ink of Rich Text: only standard text is copied
- [ ] Copy from Ink to LibreOffice of Rich Text may have spurious information (e.g also copies an unknown image.tiff name) after playing around with copy&pasting

rmottola:
Just discovered that copying 16bit image from and to PRICE&GIMP does not work.
16bit PRICE->GIMP yields a broken 8bit image
16bit GIMP->PRICE yields a correct image,  but converted to 8bit

DTW-Thalion:
Looking at how much of this sits in the backend: `xpbs.m` already covers a fair amount, which narrows down where the remaining items live.

The big-image case uses INCR incremental transfer, matching the two items marked done.

For rich text the backend has the atoms and both directions. It maps `text/rtf`, `application/rtf` and `text/richtext` to `NSRTFPboardType` when reading, and when an app asks for `NSRTFPboardType` it requests `text/rtf` from the selection owner, falling back to `application/rtf` and `text/richtext`. So "LibreOffice to Ink gives only plain text" is most likely above the backend: either the app reads `NSStringPboardType` instead of `NSRTFPboardType`, or LibreOffice advertises a target GNUstep does not list. Running gpbs with `-GNU-Debug Pbs` logs the TARGETS the owner offers, which should show which of the two it is.

The 16-bit image case is also not in the backend. `xpbs.m` only moves the raw `image/tiff` or `image/png` bytes; the 16-bit to 8-bit reduction happens where `NSBitmapImageRep` reads or writes the TIFF/PNG, so that one belongs in libs-gui rather than here.

[...]

PR #173, opened 2026-07-26
Refresh pasteboard types when another application takes the selection

DTW-Thalion:
xFixesSelectionNotify: is notified when another application takes the X selection, but it did nothing with that notification, so the pasteboard kept the types it had cached. A GNUstep application that then read the pasteboard saw the stale types and could not read the new contents. It only worked if GNUstep had previously owned the selection (which triggers xSelectionClear and a refresh) or if gpbs happened to start while the new owner already held the selection.

This re-reads the types from the new owner when the notification arrives. Two things made the direct approach (calling the existing xSelectionClear refresh from the callback) unsuitable, and both are handled here:

- Reading the new owner's types (availableTypes) does a synchronous selection round trip, which cannot run from inside the notification callback without hanging the server. The refresh is scheduled to run once the callback has returned.

- declareTypes:owner: makes pasteboardChangedOwner: take the X selection, which would take it away from the application we want to read from. A flag now tells pasteboardChangedOwner: to only update the pasteboard types in that case, without setting the selection owner.

Both of these were already noted as FIXMEs in the file.

Tested on X11: with gpbs running and no prior GNUstep copy, another application (xclip, and LibreOffice) taking the clipboard is now readable from a GNUstep pasteboard, including when the owner switches between types. A normal GNUstep copy still takes the selection and is readable by other X clients, and the server no longer takes the selection away from a foreign owner while reading from it.

Related to #45.

### DIFF
--- a/Tools/xpbs.m
+++ b/Tools/xpbs.m
@@ -254,6 +254,7 @@ NSPasteboardTypeFromAtom(Atom type)
   Time		_timeOfLastAppend;
   Time		_timeOfSetSelectionOwner;
   BOOL		_ownedByOpenStep;
+  BOOL		_updatingForeignTypes;
 }
 
 + (XPbOwner*) ownerByXPb: (Atom)p;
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):
Thank you very much. This should help a lot with the interaction with non GNUstep applications. Riccardo, Richard and I all have worked on that code over the years.

PR #174, opened 2026-07-26
Derive the default bold font from the system font family in the art and win32 enumerators

### LINKED ISSUE #170: `default{Bold,System,FixedPitch}FontName` should not hardcode font names

DTW-Thalion:
The art `FTFontEnumerator` and the Windows `WIN32FontEnumerator` kept hardcoded priority lists in `defaultBoldSystemFontName`, the same problem #172 addressed in the fontconfig enumerator: a custom system font (the `NSFont` default) without a matching `NSBoldFont` fell through to an unrelated hardcoded family, so bold UI text rendered in regular weight (issue #170, case 1).

Both backends already record each family's faces with their traits, so the bold face is now taken from the configured system font's own family, keeping the hardcoded names only as a last resort. The art enumerator also picks the fixed-pitch default from an enumerated fixed-pitch face; the win32 family enumeration does not record that trait, so its fixed-pitch default (Courier New) is unchanged.

Verified on the art backend: with `NSFont` set to a custom family, `defaultBoldSystemFontName` returns that family's bold face instead of the hardcoded one, and `defaultFixedPitchFontName` returns an enumerated fixed-pitch face instead of a fixed name; with `NSFont` unset the results are unchanged. The win32 change was built against the win32/winlib backend (MSYS2/ucrt64), where the backend compiles and links.

Together with #172 (the fontconfig enumerator) this covers the backends listed in the issue.

Closes #170.

### DIFF
--- a/Source/art/FTFontEnumerator.m
+++ b/Source/art/FTFontEnumerator.m
@@ -457,6 +457,77 @@ static void load_font_configuration(void)
   DESTROY(families_pending);
 }
 
+/* The bold face of the family that `systemFont` (a family or face name)
+   belongs to, or nil.  This lets a custom system font, set through the
+   NSFont default, resolve to its own bold variant instead of falling
+   through to an unrelated hardcoded family. */
+static NSString *
+ftBoldFaceForSystemFont(NSString *systemFont)
+{
+  NSArray *faces = [fcfg_allFontFamilies objectForKey: systemFont];
+  NSUInteger i, n;
+
+  if (faces == nil)
+    {
+      NSEnumerator *fe = [fcfg_allFontFamilies keyEnumerator];
+      NSString *family;
+
+      while (faces == nil && (family = [fe nextObject]) != nil)
+        {
+          NSArray *fs = [fcfg_allFontFamilies objectForKey: family];
+
+          n = [fs count];
+          for (i = 0; i < n; i++)
+            {
+              if ([[[fs objectAtIndex: i] objectAtIndex: 0]
+                    isEqualToString: systemFont])
+                {
+                  faces = fs;
+                  break;
+                }
+            }
+        }
+    }
+  if (faces == nil)
+    return nil;
+
+  n = [faces count];
+  for (i = 0; i < n; i++)
+    {
+      NSArray *face = [faces objectAtIndex: i];
+
+      if ([[face objectAtIndex: 3] unsignedIntValue] & NSBoldFontMask)
+        return [face objectAtIndex: 0];
+    }
+  return nil;
+}
+
+/* The name of a plain (non-bold, non-italic) enumerated face carrying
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):

probonopd:
Do we have a similar issue with monospace?

PR #177, opened 2026-07-26
Build the X11 backend without wraster by default

### LINKED ISSUE #98: wraster replacement in libs-back
@fredkiefer, 

I took an hour and went through wraster with a view of how to remove it and replace it.

On the server side, it is trivial to do - 90% of the wraster code is unused, and can be deleted.  The parts that are used in Xlib are RContext, RXImage, and RGetClosestXColor/RGBtoPixel. 6 call touchpoints. On the consumer side, there are about 39 call touchpoints. 

Cairo is the default and art doesn't use wraster at all - which is good as the fallback.

So my thought is you could move the x11 server to plain Xlib, then you could decide whether to retire the Xlib graphics backend (cleanest path), or we rewrite its 39 call sites to plain Xlib.

Thoughts?

DTW-Thalion:
Dependencies: Needs PR #175 to be merged.

This adds an `--enable-wraster` configure option and makes the X11 server and the xlib graphics backend build without wraster by default.

A new `Source/x11/xlibimage.c` reimplements the small part of the wraster image API the backend actually uses (`RContext`, `RGetClosestXColor`, and the `RXImage` create/get/put/destroy calls) on plain Xlib. It does not use MIT-SHM, and non-TrueColor visuals are resolved with `XAllocColor`. The struct and function names match `wraster.h`, so the call sites are unchanged and the graphics backend picks the implementation through a single guard in `XGPrivate.h`.

The default build (`USE_WRASTER=0`) compiles `xlibimage.c` in place of the bundled wraster sources. Configuring with `--enable-wraster` (`USE_WRASTER=1`) keeps the previous behaviour: the wraster sources are left untouched and compile exactly as before, so the two builds can be compared directly. This keeps wraster available rather than dropping it.

I checked that cairo, xlib and art build under both settings. For the xlib backend I compared the pixels read back through `GSReadRect` under each build and they are identical.

Closes #98.


fredkiefer (approved):
THis look like a great step in the direction of getting rid of wraster completely.
I suggest we leave your code in for the next release. After that  we remove support for the wraster lib and our old replacement implementation. And after that we inline your new implementation removing most of it in the process. That way wraster could be a thing of the past in two years.

PR #179, opened 2026-07-28
Draw the context shadow in GSGState rather than in each backend

DTW-Thalion:
Follows the suggestion on #168: with a -_paintPath: in every backend, the shadow drawing itself belongs in GSGState.

Each raster backend had its own copy of the same shadow rendering, differing only in how it painted the path. GSGState now does it once and calls -_paintPath: for the painting. xlib and winlib already had that method. cairo and art gain one, and their fill, even-odd fill and stroke now go through it too, which is where the rest of the removed lines come from. Net 252 lines out, 137 in.

Checked on each backend, drawing a shape with a coloured shadow at a known offset and reading the pixels back:

    x11 + cairo      Tests/cairo/dropshadow.m, 8 assertions, all pass
    x11 + art        same test, 8 assertions, all pass
    x11 + xlib       the test guards itself off here, so the same drawing was
                     run by hand against a live X server: 8 checks, all pass
    wayland + cairo  same, run under a live compositor: 8 checks, all pass
    win32 + winlib   built and run on Windows: 8 checks, all pass
    headless, opal   build clean
    xdps             not buildable here, it wants pswrap

winlib needed #180, #181, #182 and #183 before it could be checked at all: its offscreen read came back upside down, its strokes took the fill colour, compositing painted a fixed grey, and glyph bounds were empty. With those four, Tests/winlib goes from 8 failing assertions to none and the shadow drawing passes there as well. This change gives winlib the same results before and after either way.

### DIFF
--- a/Headers/gsc/GSGState.h
+++ b/Headers/gsc/GSGState.h
@@ -94,6 +94,11 @@ typedef enum {
 - (void) setShadow: (NSShadow *)shadow;
 - (NSShadow *) shadow;
 
+/* Paints the current path with the current colour. Implemented by each
+   backend, which is what lets the shadow below be drawn here. */
+- (void) _paintPath: (ctxt_object_t)drawType;
+- (void) _drawShadowForOperation: (ctxt_object_t)drawType;
+
 - (void) compositeGState: (GSGState *)source
                 fromRect: (NSRect)aRect
                  toPoint: (NSPoint)aPoint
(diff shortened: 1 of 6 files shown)


fredkiefer (approved):
Very nice thank you. It could be that this breaks the xdps backend, but it is not used anywhere. I also will need to check for the opal backend. Neither of these is release critical.

DTW-Thalion:
> Very nice thank you. It could be that this breaks the xdps backend, but it is not used anywhere. I also will need to check for the opal backend. Neither of these is release critical.

I'll let you know if it does - I am hoping to have parity between the backends established shortly.

PR #180, opened 2026-07-28
winlib: read back the pixels that were asked for

DTW-Thalion:
-GSReadRect: on winlib returned the wrong pixels in two ways.

The rows were copied in the order GetDIBits hands them back, which is bottom up, while the caller wants the top row first, so anything read back through -initWithFocusedViewRect: came out upside down. And the BitBlt took its source from the window origin rather than from the rectangle it was given, so reading part of a drawing returned whatever sat at the top of the window instead.

Tests/winlib/winlibreadrect.m covers both, drawing a red band along the bottom and a blue one along the top and then reading the whole thing and a slice of the middle. 3 of its 5 assertions fail without the change.

Built and run on Windows, win32 + winlib. Tests/winlib goes from 8 failing assertions to 4: the row order was also what failed winlibcapsjoins, winlibimage, winlibimagerotate and winlibrender. The 4 that still fail, one in win32fontmetrics and three in winlibcomposite, fail the same way before and after and are a separate matter.

Found while checking #179 on winlib, where an offscreen pixel check could not be trusted until this was sorted out.

### DIFF
--- a/Source/winlib/WIN32GState.m
+++ b/Source/winlib/WIN32GState.m
@@ -1574,6 +1574,7 @@ HBITMAP GSCreateBitmap(HDC hDC, NSInteger pixelsWide, NSInteger pixelsHigh,
   unsigned char *cdata;
   unsigned char *bits;
   int i = 0;
+  LONG row, stride;
   HDC hDC;
   HDC hdcMemDC = NULL;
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
Nice fix

PR #181, opened 2026-07-28
winlib: stroke in the stroke colour

DTW-Thalion:
The pen -setStyle: builds comes from a LOGBRUSH holding wfcolor, the fill colour, so on winlib a stroke paints in whatever colour was last filled with rather than the stroke colour. It stays hidden most of the time because -set sets both colours together; it shows as soon as they differ, as -setStroke: and the drop shadow in #168 both do.

Tests/winlib/winlibstroke.m now sets the fill green and the stroke blue and checks that a stroke paints blue and a fill paints green. The stroke assertion fails without this change.

The commented-out CreateBrushIndirect line next to it is removed, since with the LOGBRUSH now holding the stroke colour it would read as building the fill brush from the wrong one.

### DIFF
--- a/Source/winlib/WIN32GState.m
+++ b/Source/winlib/WIN32GState.m
@@ -1374,12 +1374,11 @@ HBITMAP GSCreateBitmap(HDC hDC, NSInteger pixelsWide, NSInteger pixelsHigh,
   NSInteger patternCount = 0;
   
   SetBkMode(hDC, TRANSPARENT);
+  /* The pen built from this below draws the stroke, so it takes the stroke
+     colour; the brush that fills takes the fill colour. */
   br.lbStyle = BS_SOLID;
-  br.lbColor = wfcolor;
+  br.lbColor = wscolor;
   br.lbHatch = 0;
-  /*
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
It is hard to see, why nobody ever complained here

DTW-Thalion:
> It is hard to see, why nobody ever complained here

There are a bunch of bugs I found that fall into that catagory. More to come today.

rfm:
A lot of people encounter issues and either just give up or hack in a fix or workaround that they just use themselves.  It's often much quicker/easier to make a local fix than to get a fix into a project officially, so if there is work pressure I can see why some people don't even report issues (though that doesn't necessarily stop them complaining about the bugs never getting fixed).

rfm:
A lot of people encounter issues and either just give up or hack in a fix or workaround that they just use themselves.  It's often much quicker/easier to make a local fix than to get a fix into a project officially, so if there is work pressure I can see why some people don't even report issues (though that doesn't necessarily stop them complaining about the bugs never getting fixed).

Issue #184, opened 2026-07-29
Drawing into a context from graphicsContextWithBitmapImageRep: draws nothing and crashes on flush


DTW-Thalion:
`+[NSGraphicsContext graphicsContextWithBitmapImageRep:]` hands back a context
that cannot be drawn into. Drawing is dropped and flushing it crashes.

```objc
#import <AppKit/AppKit.h>

int main(void)
{
  CREATE_AUTORELEASE_POOL(pool);
  NSBitmapImageRep *t;
  NSGraphicsContext *ctxt;

  [NSApplication sharedApplication];

  t = [[NSBitmapImageRep alloc]
    initWithBitmapDataPlanes: NULL pixelsWide: 8 pixelsHigh: 8
    bitsPerSample: 8 samplesPerPixel: 4 hasAlpha: YES isPlanar: NO
    colorSpaceName: NSDeviceRGBColorSpace bytesPerRow: 0 bitsPerPixel: 0];
  ctxt = [NSGraphicsContext graphicsContextWithBitmapImageRep: t];

  [NSGraphicsContext saveGraphicsState];
  [NSGraphicsContext setCurrentContext: ctxt];
  [[NSColor redColor] set];
  NSRectFill(NSMakeRect(0, 0, 8, 8));
  [ctxt flushGraphics];
  [NSGraphicsContext restoreGraphicsState];

  RELEASE(pool);
  return 0;
}
```

A CairoContext comes back, so the caller has no way to tell anything is wrong.
The rect fill leaves the bitmap untouched, and `-flushGraphics` segfaults:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff3ae2870 in XFlush () from /lib/x86_64-linux-gnu/libX11.so.6
#0  XFlush ()
#1  main () at bmp2.m:24
```

Drawing an image into the same context instead of filling a rect reports the
missing surface rather than crashing:

```
WARNING: -drawGState called with a NULL target context ((null)) or source context (0x...)
```

The context has no surface because the destination attribute is only acted on
when it is a window. From `-[GSContext initWithContextInfo:]`
(Source/gsc/GSContext.m):

```objc
      // Special handling for window drawing
      dest = [info objectForKey: NSGraphicsContextDestinationAttributeName];
      if ((dest != nil) && [dest isKindOfClass: [NSWindow class]])
        {
          server = GSCurrentServer();
          [server setWindowdevice: [(NSWindow*)dest windowNumber]
                  forContext: self];
[...]

PR #185, opened 2026-07-29
Fix: do not flush a context that has no window server

DTW-Thalion:
A context from `+[NSGraphicsContext graphicsContextWithBitmapImageRep:]`
segfaults on its first flush. `-[CairoContext flushGraphics]` passes the server
of the context straight to `XFlush`, and `server` is only assigned in
`-[GSContext initWithContextInfo:]` when the destination attribute is a window,
so any other destination reaches `XFlush(NULL)`.

Do not flush when there is no server.

This only stops the crash. Drawing into such a context still produces nothing,
because the cairo backend has no surface class to attach to a bitmap
destination: every CairoSurface subclass is backed by a window or by a print
stream. That part is #184 and is not addressed here.

Tests/cairo/bitmapcontextflush.m aborted at the flush before, with 1 of its 2
assertions run, and both pass after. The cairo suite goes from 319 to 321
assertions with no failures, and the whole backend suite runs clean on
x11/cairo.

Refers to #184

### DIFF
--- a/Tests/cairo/bitmapcontextflush.m
+++ b/Tests/cairo/bitmapcontextflush.m
@@ -1,0 +1,71 @@
+/* A graphics context whose destination is an NSBitmapImageRep is not attached
+ * to a window, so it has no window server behind it.  Flushing such a context
+ * has to be harmless: the x11 flush reaches for the server of the context, and
+ * with no window there is nothing to ask.
+ *
+ * It needs a running window server to create the context at all, so it skips
+ * cleanly when there is none, and guards on the cairo graphics backend.
+ */
+#import <Foundation/NSObject.h>
+#import "Testing.h"
+#include "config.h"
+
+#if defined(BUILD_GRAPHICS) && defined(GRAPHICS_cairo) \
+  && BUILD_GRAPHICS == GRAPHICS_cairo
+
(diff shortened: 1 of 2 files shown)


Tests/cairo/bitmapcontextflush.m
```diff
@@ -0,0 +1,71 @@
...
+#if defined(BUILD_GRAPHICS) && defined(GRAPHICS_cairo) \
+  && BUILD_GRAPHICS == GRAPHICS_cairo
+
+#import <AppKit/AppKit.h>
+#include <stdlib.h>
+
+int
+main(int argc, const char **argv)
+{
+  CREATE_AUTORELEASE_POOL(pool);
```
fredkiefer, line 22: Didn't you switch to `START_SET`?

fredkiefer (approved):

Tests/cairo/bitmapcontextflush.m
DTW-Thalion: Yes, there are about 30 of these that will get cleaned up in another hygiene PR - I needed to work through a few waves of memory management issues, but I had to wait until things were merged in before I could fix (or end up making a lot more work for both of us).

Issue #167, opened 2026-07-15
Test placement: general tests in gui, backend-specific tests kept in libs-back


DTW-Thalion:
@fredkiefer  - Just going through to address the test coverage now.

Over the backend test-coverage PRs (#86 onward) a number of tests ended up under one backend's Tests directory and guarded to that backend, while (as you pointed out) what they actually check is general behavior that every backend has to satisfy. I am about  to do some PR hygiene, but before moving or adding anything further I would like to make sure I have a consistent picture and apply it in one pass rather than case by case.

Proposed rule:

1. A test that exercises code specific to one backend stays where it is, guarded to that backend. That covers the WGL and EGL context glue, the wayland wl_shm surface, the cairo PDF/PS and surface/device code, isCompatibleBitmap:, the fontconfig pattern and installer, and the wraster/x11 tests.

2. A test that exercises general AppKit drawing or font behavior moves to libs-gui, so it runs against whichever backend is installed instead of only one. That covers compositing, fill, stroke, caps and joins, curves, the winding rule, clip stacking, gsave/restore, gradients, image drawing, alpha and rotation, text drawing, and font enumeration and metrics. Several of these exist today as near-duplicate copies under art, cairo and winlib and would collapse into a single test each in gui.

3. A test of the GSDisplayServer contract that every server implements the same way (window creation, bounds and level, screen queries, mouse location) becomes one generic test in libs-back, run against the built server, instead of a per-server copy.

One test I would instead drop than move is the win32 WS_ style mapping (#142). It checks a private mapping with no cross-backend meaning, which matches your comment there.

The tradeoff you already noted on #99 is that a gui-hosted rendering test only runs where the CI or the user has a drawable backend, so on a headless gui build it skips instead of failing. I think that is still the right home for it.

If this matches what you have in mind I will do it as a few grouped PRs: add the general tests to gui, verified on cairo first, then a libs-back pass to remove the moved copies and add the generic server-contract tests. Open PRs #144, #152, #154 and #157 would be closed and refiled in the right place; #153, #155 and #156 stay backend-specific.

Are there any you would classify differently?

fredkiefer:
This sounds excellent to me. One thing we could discuss is whether it would be worthwhile to run the gui tests even when building back in the CI? That way we could address the issue you mention that many gui tests only get executed locally when back is complied first. At the moment building of back is disabled for the gui CI. If this is possible we have the tests in the right place and still have a broad coverage for many backends.

[...]

PR #190, opened 2026-07-30
Fix: draw a gradient on every backend, not just cairo and opal

### LINKED ISSUE #187: NSGradient does not draw on the art, xlib or winlib backends, and the raise hangs the process
`-[NSGradient drawInRect:angle:]` reaches the graphics state through `-drawGradient:fromPoint:toPoint:options:`, and the radial calls go through `-drawGradient:fromCenter:radius:toCenter:radius:options:`. Both are declared in `Headers/gsc/GSGStateOps.h` and raise `subclassResponsibility` in `Source/gsc/GSGState.m`, at lines 1429 and 1419.

Only `CairoGState` and `OpalGState` override them. `ARTGState`, `XGGState`, `WIN32GState`, `Win32CairoGState`, `HeadlessGState` and `GSStreamGState` do not, so no gradient draws on any of those backends.

The raise does not surface as an error. It is uncaught, so `_NSAppKitUncaughtExceptionHandler` opens a modal alert panel and `-[NSApplication runModalForWindow:]` waits for a button press. With no window manager, or under a headless server, that press never comes and the process blocks for good. A backtrace taken from a run under Xvfb with the xlib backend:

```
#1  -[GSRunLoopCtxtUnix pollUntil:within:]        GSRunLoopCtxt.m:360
#2  -[NSRunLoop acceptInputForMode:beforeDate:]   NSRunLoop.m:1337
#4  -[GSDisplayServer(EventOps) getEventMatchingMask:beforeDate:inMode:dequeue:]
#7  -[NSApplication runModalForWindow:]           NSApplication.m:1829
#8  -[GSAlertPanel runModal]                      NSAlert.m:812
#9  GSRunExceptionPanel                           NSAlert.m:2183
#10 _NSAppKitUncaughtExceptionHandler             NSApplication.m:136
#16 -[NSObject(GNUstepBase) subclassResponsibility:]
#17 -[NSGradient drawInRect:angle:]               NSGradient.m:109
```

To reproduce, against a backend configured `--enable-graphics=art` or `--enable-graphics=xlib`:

```objc
#import <AppKit/AppKit.h>

int main(void)
{
  NSAutoreleasePool *p = [NSAutoreleasePool new];
  NSImage *img;
  NSGradient *g;

  [NSApplication sharedApplication];
  img = [[NSImage alloc] initWithSize: NSMakeSize(32, 32)];
  [img lockFocus];
  g = [[NSGradient alloc] initWithStartingColor: [NSColor blackColor]
                                    endingColor: [NSColor whiteColor]];
  printf("about to draw\n"); fflush(stdout);
  [g drawInRect: NSMakeRect(0, 0, 32, 32) angle: 0.0];
  printf("returned\n"); fflush(stdout);
  [g release];
  [img unlockFocus];
  [p release];
  return 0;
}
```

It prints `about to draw` and then hangs. The same program prints `returned` on a cairo build.

Two consequences worth noting. A gradient is a plain AppKit drawing call, so an application that uses one works on cairo and stops responding on art or xlib with no message. And a test that draws a gradient does not fail, it blocks, which `gnustep-tests --timeout` does not catch, because the test program ignores SIGTERM and plain `timeout` never kills it.

A fix in each backend would mean writing gradient rasterisation three times. `GSGState` already has the colour, clip and fill primitives needed to interpolate the stops and paint the result, so a single implementation there would cover art, xlib, winlib, headless and the PostScript stream context at once, with cairo and opal keeping their native versions.

DTW-Thalion:
An NSGradient did not draw on the art, xlib or winlib backends. GSGState raises subclassResponsibility for -drawGradient:fromPoint:toPoint:options: and its radial counterpart, and only CairoGState and OpalGState override them. The raise was uncaught, so AppKit opened a modal alert panel and the process waited on it for good, which is why this shows up as a hang rather than an error.

GSGState now paints a gradient itself, interpolating the stops and filling a series of bands with the fill operation each backend already has. Cairo and opal keep their native versions. A stop whose colour has no RGB components, a pattern for instance, is skipped instead of raising.

Bands are a device pixel apart up to a ceiling of 256, so a very large gradient is approximated, and they reach past the area painted because the caller clips first. Verified on cairo, art and xlib; winlib and headless get the same code but were not exercised.

Tests/cairo/gradient.m now builds against whichever backend is built. Where the general drawing tests should live is #167.

Its 2 assertions fail before and pass after, on art and xlib. The full suite is unchanged at 319 passed, 24 skipped, 0 failed.

Closes #187.

### DIFF
--- a/Source/gsc/GSGState.m
+++ b/Source/gsc/GSGState.m
@@ -25,6 +25,7 @@
 */
 
 #include "config.h"
+#import <Foundation/NSException.h>
 #import <Foundation/NSObjCRuntime.h>
 #import <Foundation/NSValue.h>
 #import <Foundation/NSDictionary.h>
@@ -32,6 +33,7 @@
 #import <AppKit/NSBezierPath.h>
 #import <AppKit/NSColor.h>
 #import <AppKit/NSColorSpace.h>
+#import <AppKit/NSGradient.h>
 #import <AppKit/NSImage.h>
 #import <GNUstepGUI/GSFontInfo.h>
 #import <AppKit/NSGraphics.h>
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):
I am not sure whether the circular gradient is fully correct, but a lot better than failing with an exception.
Thank you!

PR #191, opened 2026-07-30
tests: let the test framework manage the autorelease pools

DTW-Thalion:
Tests here manage their own autorelease pools. Some open one inside a START_SET, which already makes one; most of the rest have no set at all, so an exception leaves the pool unreleased and the checks that give up early do so with a return.

The redundant inner pools are removed. The tests with no set are wrapped in one and now skip instead of returning, since a return between START_SET and END_SET jumps past the end of the set. That includes the case where the backend cannot start, which used to abort the whole file. The four tests with neither a pool nor a set get one, so they report why they did nothing.

Tests/cairo/gradient.m is left alone because #190 changes it. Two sets could already be left open this way, in x11/convert.m and x11/rcontext.m, so those returns become skips too.

Run on x11 with cairo, art and xlib, on wayland with cairo, and on win32 with winlib, with per-file assertion counts unchanged everywhere: 319 passed on x11 with cairo before and after, 99 passed and 8 failed on winlib before and after. Only skipped sets rise, 24 to 28 on x11 with cairo and 18 to 46 on winlib, being tests that previously reported nothing.

### DIFF
--- a/Tests/art/compositedest.m
+++ b/Tests/art/compositedest.m
@@ -77,6 +77,8 @@ blend2(int a, int wa, int b, int wb)
 int
 main(void)
 {
+  START_SET("art destination compositing")
+
   unsigned char src[3], dst[3], srca[1], dsta[1];
   composite_run_t c;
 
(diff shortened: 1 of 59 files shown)


fredkiefer (approved):
Thank you, this new pattern is a lot cleaner.

PR #202, opened 2026-07-31
Fix: clip to a path of more than one subpath as one region on the xlib backend

### LINKED ISSUE #189: xlib: an even-odd clip admits the wrong region
`Tests/cairo/windingrule.m` and `Tests/cairo/clipsaverestore.m`, run against an xlib build with their guard widened past cairo, each fail exactly the assertions that use an even-odd clip and pass everything else.

```
Failed test: windingrule.m:129 ... an even-odd clip admits drawing in the outer ring
Failed test: windingrule.m:131 ... an even-odd clip excludes the inner hole
Failed test: clipsaverestore.m:152 ... an even-odd ring clip still paints the ring after save/restore
Failed test: clipsaverestore.m:154 ... an even-odd ring clip keeps its hole after save/restore
```

The even-odd and non-zero fill assertions in `windingrule.m` pass, so the winding rule itself is applied correctly when filling a path. The plain rectangular and oval clips in `clipsaverestore.m` pass, and so does the clip stack across a save and restore. What fails is the even-odd rule applied to a clip, which points at `-DPSeoclip` in `XGGState` rather than at the fill or the clip stack.

The cairo and art backends pass all four.

These are only reachable now that #187 is fixed. Both files draw a gradient earlier on, which blocked the process before it could get this far.

Reproduce with a backend configured `--enable-graphics=xlib`, and the guard in those two files widened from `BUILD_GRAPHICS == GRAPHICS_cairo` to `BUILD_GRAPHICS != GRAPHICS_headless`.

DTW-Thalion:
Clipping to a path on the xlib backend emitted each subpath as it was reached and intersected the result with the clip already in place. A ring was therefore clipped to its outer boundary and then to its inner one, so what the clip admitted was the hole rather than the ring, and any drawing in the ring was lost.

A clip is now gathered the way a fill already is, and the subpaths are combined into one region before it is applied: an even-odd clip takes their symmetric difference and a non-zero clip their union, since X has no polygon with a hole in it. A stroke still goes out subpath by subpath, as it draws rather than building a region.

The union stands in for a non-zero clip whose subpaths wind in opposite directions, which region algebra cannot express. Nothing in these tests covers that, and it is closer than the intersection was.

The tests are Tests/gui/NSGraphicsContext/clipping.m and Tests/gui/NSBezierPath/drawingWindingRule.m in libs-gui, run against a backend configured --enable-graphics=xlib.

Four assertions fail before the change and none after. The five gui drawing and font directories go from 965 passed and 14 failed to 969 and 10, with no other assertion moving, and this repository's own suite is unchanged.

Closes #189

### DIFF
--- a/Source/xlib/XGGState.m
+++ b/Source/xlib/XGGState.m
@@ -922,24 +922,88 @@ static Region emptyRegion;
     }
 }
 
+/* Clip to a path of more than one subpath. X has no polygon with a hole in it,
+   so each subpath becomes a region of its own and they are combined: an
+   even-odd clip takes their symmetric difference, a non-zero clip their union.
+   Clipping to each subpath in turn would intersect them, which leaves a ring
+   admitting only its hole. */
+- (void) _doComplexClip: (XPoint*)pts
+                       : (int*)types
+                       : (int)count
+                   draw: (ctxt_object_t)type
+{
+  Region total = XCreateRegion();
+  int    start = 0;
+  int    i;
+
+  for (i = 1; i <= count; i++)
+    {
+      if (i == count || types[i] == 0)
+        {
+          int n = i - start;
+
+          if (n > 2)
+            {
+              Region sub = XPolygonRegion(&pts[start], n,
+                (type == path_eoclip) ? EvenOddRule : WindingRule);
+              Region combined = XCreateRegion();
+
+              if (type == path_eoclip)
+                {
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):
I must admit I never did fully understand this code. I understand your changes and see, why they should yield better results, but it won't help me understand the rest.

PR #203, opened 2026-07-31
Fix: font size, string width and glyph bounding box on the xlib backend

DTW-Thalion:
On the xlib backend a font reported the metrics of whatever size it had been enumerated at rather than the size it was asked for, so an ascender, a glyph advance and a maximum advancement were the same at 28 points as at 14. The width of a string was the width of the ink it puts down rather than how far it advances, so four W's measured less than four single W's. The bounding rectangle of a glyph was as wide as the glyph advances, so an i had no visible side bearings.

The pattern a font is matched from already carries a size, and fontconfig appends a value rather than replacing it, so the earlier size is removed before the requested one is added. A string is measured by its advance. The bounding rectangle comes from the face, in the same way the bezier path for a glyph already does, because the extents Xft reports are those of the padded image it renders.

The tests are Tests/gui/NSFont/fontmetrics.m and glyphmetrics.m in libs-gui, run against a backend configured --enable-graphics=xlib.

Five assertions fail before the change and none after. The five gui drawing and font directories go from 965 passed and 14 failed to 970 and 9, with no other assertion moving, and this repository's own suite is unchanged.

### DIFF
--- a/Source/xlib/GSXftFontInfo.m
+++ b/Source/xlib/GSXftFontInfo.m
@@ -115,11 +115,13 @@
   [string getCharacters: (unichar*)str];
   XftTextExtents16 ([XGServer xDisplay],
 		    font_info,
-		    str, 
+		    str,
 		    len,
 		    &extents);
 
-  return extents.width;
+  /* The width of a string is how far it advances the drawing position, not
+     the width of the ink it puts down. */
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):
Good catch

PR #208, opened 2026-08-01
Fix: do not probe for window offsets when no window manager is running

### LINKED ISSUE #207: x11: with no window manager every process spends fifteen seconds failing to determine window offsets
On a display with no window manager, every process using the X11 server spends about fifteen seconds working out window decoration offsets that cannot be worked out, discards the result, and the next process repeats it. On the xlib backend the whole of libs-gui's `Tests/gui` takes 88 minutes where it takes about a minute and a half on cairo, and the difference is entirely this.

`-_checkStyle:` creates a probe window for each of the fifteen window styles. `GSBackChecksOffsetsOnScreen` defaults to NO, and when it is NO the probe is placed deliberately off screen at x = -200. The loop that follows waits for a `VisibilityNotify` saying the window is visible. A window entirely off screen is fully obscured, so the server correctly reports `VisibilityFullyObscured`, `window->visibility` stays above 1, and the loop sleeps for its full second before giving up. Fifteen styles, fifteen seconds.

The other way out of that loop is a `ReparentNotify`, which only a window manager sends, so with none running neither exit condition can be met.

It never amortises because the offsets are cached in the `_GNUSTEP_FRAME_OFFSETS` property on the root window only when all fifteen styles succeed. With no window manager none of them do, so nothing is ever written and every process starts again.

The result is discarded in any case. With no window manager `-styleoffsets:::::` finds `known` still NO for every style and falls through to its "no known WM protocols" branch, which answers zero offsets. That is the same answer the probe would have given had it worked.

Only the timing differs between backends, not the outcome. The probe window takes its depth from the graphics backend: cairo's is 32 bits deep and the server reports `VisibilityUnobscured` for it, so the loop exits at once; xlib's is 24 bits deep and is reported fully obscured. Both then fail to determine any offsets. art behaves as xlib does, opal as cairo does.

Measured on `Tests/gui/NSTableView` with the property cleared first, same machine, same Xvfb, only the backend bundle swapped:

```
cairo    1s    0 one-second waits
art    153s  150 one-second waits
xlib   154s  150 one-second waits
opal     1s    0 one-second waits
```

Reproduce with an X server and no window manager, for example `Xvfb :99`, a backend configured `--enable-graphics=xlib`, and `xprop -root -remove _GNUSTEP_FRAME_OFFSETS` before the run so the cache starts empty.

DTW-Thalion:
With no window manager running, each of the fifteen window styles is probed with a window placed off screen, which can never become visible and is never reparented, so the loop waiting on it spends a full second and learns nothing. Nothing is cached either, because the offsets are only written to the root window when every style succeeds, so every process repeats the whole thing.

A new `-_noWindowManager` answers whether anything else holds SubstructureRedirect on the root window, which a window manager always does. When there is none the offsets are set to zero, which is what the code already falls back to, and the probe is skipped. They are deliberately not written to the root window, where a window manager started later would find them and use them in place of its own.

Nothing changes when a window manager is running: with the cached property cleared first, the offsets are identical under openbox, metacity and twm.

There is no test, because the behaviour only differs when no window manager is running and the values produced are the ones the fallback already gave.

Tests/gui in libs-gui, against a backend configured --enable-graphics=xlib with no window manager, goes from 88m 17s to 1m 28s, with 4459 passed and 8 failed either way.

Closes #207

### DIFF
--- a/Source/x11/XGServerWindow.m
+++ b/Source/x11/XGServerWindow.m
@@ -791,6 +791,53 @@ _get_next_prop_new_event(Display *display, XEvent *event, char *arg)
   return extents;
 }
 
+static BOOL	wmTestFailed;
+
+static int
+_wmTestErrorHandler(Display *display, XErrorEvent *event)
+{
+  if (event->error_code == BadAccess)
+    {
+      wmTestFailed = YES;
+    }
+  return 0;
+}
+
+/* Is there no window manager running at all?  Only one client at a time may
+   select SubstructureRedirect on the root window and a window manager always
+   does, so being allowed to select it means there is none.  It is given back
+   immediately.  The answer is worked out once.
+*/
+- (BOOL) _noWindowManager
+{
+  static BOOL	checked = NO;
+  static BOOL	none = NO;
+
+  if (checked == NO)
+    {
+      Window		root = DefaultRootWindow(dpy);
+      XWindowAttributes	attributes;
+      int		(*previous)(Display *, XErrorEvent *);
+
+      checked = YES;
+      wmTestFailed = NO;
+      XGetWindowAttributes(dpy, root, &attributes);
+      XSync(dpy, False);
+      previous = XSetErrorHandler(_wmTestErrorHandler);
+      XSelectInput(dpy, root,
+                   attributes.your_event_mask | SubstructureRedirectMask);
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):
Good catch, and really important for the CI tests

PR #218, opened 2026-08-05
Fix: give a context from graphicsContextWithBitmapImageRep: a cairo surface

### LINKED ISSUE #184: Drawing into a context from graphicsContextWithBitmapImageRep: draws nothing and crashes on flush

DTW-Thalion:
`+[NSGraphicsContext graphicsContextWithBitmapImageRep:]` returns a CairoContext
with no surface, because `-[GSContext initWithContextInfo:]` only acts on a
window destination. Anything drawn into that context is dropped.

CairoBitmapSurface draws into a cairo image surface and writes the pixels back
to the representation on `-flushGraphics`. It reads the representation first, so
a second context for the same bitmap adds to what is already there. Cairo holds
a pixel as a native word with the alpha in the top byte, the representation as
four bytes with the alpha last.

Only a 32 bit non planar bitmap in device or calibrated RGB with premultiplied
alpha last gets a surface; any other layout is as it was. On macOS the drawing
reaches the representation without a flush; here it needs one.

Tests/cairo/bitmapcontextdraw.m, run under a display with
`gnustep-tests Tests/cairo/bitmapcontextdraw.m`.

That file fails 7 of its 10 assertions against master and passes 10 here. Tests/
goes from 255 passed 0 failed to 265 passed 0 failed, with the same 31 skipped
sets.

Closes #184

### DIFF
--- a/Headers/cairo/CairoBitmapSurface.h
+++ b/Headers/cairo/CairoBitmapSurface.h
@@ -1,0 +1,47 @@
+/*
+   Copyright (C) 2026 Free Software Foundation, Inc.
+
+   Author: Todd White <todd.white@thalion.global>
+   Date: August 2026
+
+   This file is part of GNUstep.
+
+   This library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Lesser General Public
+   License as published by the Free Software Foundation; either
+   version 2 of the License, or (at your option) any later version.
+
+   This library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the GNU
+   Lesser General Public License for more details.
+
+   You should have received a copy of the GNU Lesser General Public
+   License along with this library; see the file COPYING.LIB.
+   If not, see <http://www.gnu.org/licenses/> or write to the
+   Free Software Foundation, 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110-1301, USA.
+*/
+
+#ifndef CairoBitmapSurface_h
+#define CairoBitmapSurface_h
+
+#include "cairo/CairoSurface.h"
+
+@class NSBitmapImageRep;
+
+/* A surface that draws into the bytes of an NSBitmapImageRep.  Cairo keeps
+ * its own buffer, in its own pixel layout, and the drawing is written back to
+ * the representation when the surface is flushed.
+ */
+@interface CairoBitmapSurface : CairoSurface
+{
+  NSBitmapImageRep *_rep;
+}
+
+/* Whether a representation is in a layout this surface can write back to. */
++ (BOOL) handlesBitmap: (NSBitmapImageRep *)rep;
+
+@end
+
+#endif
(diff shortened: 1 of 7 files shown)


fredkiefer (approved):
Excellent! Thank you very much

PR #224, opened 2026-08-06
Fix: read the GDI pitch flag the way GDI defines it

DTW-Thalion:
A font reports the wrong pitch on winlib. WIN32FontInfo reads the GDI text
metric as `isFixedPitch = TMPF_FIXED_PITCH & metric.tmPitchAndFamily`, and that
bit is set for a font of VARIABLE pitch, the opposite of what its name says.

Measured at 14pt with GetCharABCWidths: Courier New and Consolas report the bit
clear and advance W and i alike, while Tahoma and Arial report it set and
advance them 14 against 2. So Courier New answered NO to the isFixedPitch
question and Tahoma answered YES.

The test asks the font itself rather than naming a face: a font whose W and i
advance alike is fixed pitch, and the reported flag has to agree with that.

What it fixes beyond the flag: NSSecureTextFieldCell substitutes the user
fixed pitch font when it is handed a variable font, and decides that from the
same flag, so on Windows it substituted nothing. Tests/gui/NSSecureTextField
in libs-gui goes from 14 passed 1 failed to 15 passed 0 failed with this.

Tests/winlib/fixedpitch.m is new. The winlib set is 52 passed 2 failed against
master and 54 passed 0 failed here. No CI job builds winlib, so the checks
here do not exercise any of it.

### DIFF
--- a/Source/winlib/WIN32FontInfo.m
+++ b/Source/winlib/WIN32FontInfo.m
@@ -526,7 +526,11 @@ NSLog(@"No glyph for U%d", c);
   DeleteDC(hdc);
 
   // Fill the ivars
-  isFixedPitch = TMPF_FIXED_PITCH & metric.tmPitchAndFamily;
+  /* TMPF_FIXED_PITCH is set for a font of VARIABLE pitch, which is the
+     opposite of what the name says. Courier New and Consolas report the bit
+     clear and advance W and i alike; Tahoma and Arial report it set and do
+     not. */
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
Good catch. In this case I really blame Microsoft not us for the bug.

DTW-Thalion:
> Good catch. In this case I really blame Microsoft not us for the bug.

I agree with this assessment. 😄

PR #225, opened 2026-08-06
Fix: enumerate the fonts fontconfig knows about on art

### LINKED ISSUE #206: art fonts: userFixedPitchFontOfSize: returns a proportional font because only one family is enumerated
`Tests/gui/NSFont/glyphmetrics.m` in libs-gui fails its last assertion, that a fixed pitch font advances i and W the same, on a `--enable-graphics=art` build. Every other backend passes it.

`[NSFont userFixedPitchFontOfSize:]` returns Helvetica there, and at 24 points its i advances 5.33 and its W 22.66. Two lines appear on the console before that:

```
Font not found Courier
The font specified for NSUserFixedPitchFont, Courier, can't be found.
```

The reason is not that the wrong font is chosen. There is no fixed pitch font to choose. `Source/art/GNUmakefile` is the only one that builds `FTFontEnumerator.m`; `Source/cairo/GNUmakefile`, `Source/opal/GNUmakefile` and `Source/xlib/GNUmakefile` all build `../fontconfig/FCFontEnumerator.m` instead. `load_font_configuration` reads only `.nfont` and `.font` bundles under the Library Fonts directories, and this repository installs exactly one, `Fonts/Helvetica.nfont`.

Asking `NSFontManager` for `availableFontFamilies` on the same machine gives 61 families under cairo, seven of which have a fixed pitch face, and 1 family under art.

So `ftPlainFaceWithTrait(NSFixedPitchFontMask)` finds nothing, the `BitstreamVeraSansMono-Roman` and `FreeMono` fallbacks in `defaultFixedPitchFontName` are not installed either, and it returns `@"Courier"`, which is not among the enumerated fonts. NSFont reports that and falls back to the system font, which is the Helvetica the test then measures.

The same applies to `defaultSystemFontName`: art reports Helvetica where the fontconfig backends report DejaVu Sans. It shows up as a font metrics difference rather than a failure, so only the fixed pitch case is caught.

There are two ways out: One is to install a fixed pitch bundle beside `Fonts/Helvetica.nfont`, whose `FontInfo.plist` gives its foundry as URW; the same URW set carries a monospaced face, which the fontconfig backends enumerate here as `NimbusMonoPS-Regular`. The other is to give art the fontconfig enumerator the other three backends use, so that it sees the fonts the system actually has.

Reproduce with a backend configured `--enable-graphics=art`, then run `Tests/gui/NSFont` from libs-gui.

Is either of these wanted, and if so which?

DTW-Thalion:
art enumerates the .nfont and .font bundles under the Library Fonts
directories, and this repository installs just the Helvetica bundle, so art
sees a single family. With no fixed pitch face to answer with,
userFixedPitchFontOfSize: falls back to the system font: at 24pt its i
advances 5 and its W 23.

art now uses the fontconfig enumerator that cairo, opal and xlib use, so it
sees what the system has. FTFontEnumerator becomes a subclass of
FCFontEnumerator, and FTFaceInfo a subclass of FCFaceInfo that reads the file,
the face index and the hinting out of the pattern fontconfig matched. That
drops the private font scanning, 570 lines of it, and the renderer takes its
hinting from the desktop settings rather than from a plist. The FreeType cache
keys on the face, so a face index other than 0 now opens correctly.

Measured here, art goes from 1 family with no fixed pitch face to 61 with 7,
the system font from Helvetica to DejaVuSans, and the fixed pitch font to
DejaVuSansMono, whose i and W both advance 14.

Tests/art/fixedpitch.m is new. The art set is 32 passed 3 failed against
master and 35 passed 0 failed here.

Closes #206


fredkiefer:
I am unsure about this PR. It is great to remove so much code, but now the art backend would depend on FontConfig. Up to now the art backend was a great option for old machines where there was no suitable FontConfig implementation available. That is computers where cairo, our preferred backend, was not possible.

DTW-Thalion:
I am fine with that too - this was done for parity amongst the backends to try to ensure upstream compatibility at the application level, but I can certainly see the use case of art as a "backend of last resort" if you will. 

We can let people think about it for a bit, and then if not desired, we can close it and move on.

rmottola:
I would keep this open right now and have the time to test it, I think there is no hurry with the art backend anyway.
I found it nice that art could use nfont packages, albeit a little bit foreign on Linux, very OpenStep/Mac like. Also with remote display export it uses host and not server fonts. Interesting alternative.

PR #226, opened 2026-08-06
Fix: flip the _NET_WORKAREA origin and count monitors, not outputs

DTW-Thalion:
_NET_WORKAREA is used as the screen frame for a single monitor. Two things are
wrong with that.

The property is in X coordinates, where y grows downwards, and the origin was
set to zero rather than flipped. A panel at the top and one at the bottom
reserve the same rows and report the same height, so only the origin tells
them apart. Zeroing it suits a panel at the top and is wrong by the height of
the panel for one at the bottom.

The single monitor test read monitorsCount while it still held
screen_res->noutput, so on a driver reporting an output per connector the work
area was not used.

Measured on a 1920x1080 screen with 40 rows reserved at the bottom: a bar
there was mapped at X row 1030, over a panel at rows 1040 to 1079, and is now
at 990. On the Xorg dummy driver, with 16 outputs, the frame was 1920x1080 and
is now 1920x1040 at y 40.

Tests/x11 gives 259 passed 0 failed against 257 passed 2 failed, and two
monitors are unchanged. The count needs several outputs, which CI has not.

### DIFF
--- a/Source/x11/XGServerWindow.m
+++ b/Source/x11/XGServerWindow.m
@@ -4752,11 +4752,7 @@ _workAreas(Display *dpy, Window root, CGFloat screenHeight)
       frame.origin.y = (uint32_t)*items++;
       frame.size.width = (uint32_t)*items++;
       frame.size.height = (uint32_t)*items++;
-      /* X coordinates need to be flipped to OpenStep coordinates.  The origin
-       * says which end of the screen the reserved rows are at: a panel at the
-       * top and one at the bottom reserve the same number of rows and so give
-       * the same height, and only the origin tells them apart.
-       */
+      // X coordinates need to be flipped to OpenStep coordinates
       frame.origin.y = screenHeight - frame.origin.y - frame.size.height;
       [array addObject: [NSValue valueWithRect: frame]];
     }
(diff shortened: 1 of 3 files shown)


probonopd:
Confirming both halves of this from a desktop that reserves space at the top and at the bottom, and adding what the origin means for the code that is about to consume the work area.

With a 22 pixel menu bar at the top and a dock at the bottom of an 800x600 screen, the window manager publishes `_NET_WORKAREA = 0, 22, 800, 450`. Read as this PR does, that is `0 128 800 450` in screen coordinates: 22 reserved at the top, 128 at the bottom. Read as master does, with the origin zeroed, it becomes `0 0 800 450`, which says the bottom 128 pixels are free and the top 150 are not. The two readings differ by 128 pixels in the opposite direction, and only the first one describes the screen.

That matters beyond the screen frame. gnustep/libs-gui#953 builds `-[NSScreen visibleFrame]` from the work area, so with the origin zeroed an application would be told to keep out of the wrong edge - it would avoid a menu bar that is not there and place its windows under the dock.

The output count is the same story here as on the dummy driver: an ordinary laptop lists `eDP-1` plus five disconnected connectors, so `screen_res->noutput` is 6 and the single monitor test never fires, while the same machine under Xvfb reports one output and behaves differently. Counting outputs that have a CRTC makes the two agree.

We carry both fixes downstream and have seen no fallout from either.

PR #233, opened 2026-08-09
Skip on the absence of a display server, not of DISPLAY

DTW-Thalion:
14 tests in Tests/cairo and Tests/gsc skip when the DISPLAY environment
variable is unset. None of them is an X11 test: 6 exercise the cairo graphics
backend and 8 the shared gsc code. A backend that is not X11 has no DISPLAY,
so on such a backend the set skips, reports success and asserts nothing.

They now ask whether a display server can be reached, which is what
Tests/server/servercontract.m already does. GSCurrentServer() is declared in
GSDisplayServer.h, so that header is imported where it was not there already.
The file comments that described the DISPLAY check are updated. Nothing
outside Tests changes.

Run gnustep-tests Tests/cairo Tests/gsc on a cairo build, once with a display
and once without.

Under X11 the counts are the same as before, with a display and without one.
On a backend that has no DISPLAY all 14 go from skipping to running, and 94
assertions pass with none failing.

### DIFF
--- a/Tests/cairo/bitmapcontextdraw.m
+++ b/Tests/cairo/bitmapcontextdraw.m
@@ -17,6 +17,7 @@
   && BUILD_GRAPHICS == GRAPHICS_cairo
 
 #import <AppKit/AppKit.h>
+#import <GNUstepGUI/GSDisplayServer.h>
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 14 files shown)


Tests/cairo/bitmapcontextdraw.m
```diff
@@ -85,14 +86,16 @@
   int x;
   int y;
 
-  if (getenv("DISPLAY") == NULL || *getenv("DISPLAY") == '\0')
-    {
-      SKIP("no window server available")
-    }
-
+  /* Ask whether a display server can be reached rather than whether X11's
```
fredkiefer: This comment explains the change but not the remaining code. As with many AI generated comments, the code would be better off without it.

fredkiefer (changes requested):
Please remove half of the new comments. Otherwise this change is great

DTW-Thalion:
Removed in all 14 files. The remaining comment change is in the file headers, which said the test opens the display named by the environment; that is no longer what the code does, so the sentence is reworded.

fredkiefer (approved):

PR #234, opened 2026-08-09
Add an android display server drawing through cairo

DTW-Thalion:
This adds an android server over the existing cairo graphics path, on the
Wayland pattern. 

AndroidServer is a GSDisplayServer subclass holding a window
table, screen metrics and window ordering; AndroidCairoSurface is a
CairoSurface subclass. 

configure takes android in --enable-server and errors,
as wayland does, rather than falling back to xlib when cairo or freetype is
missing, and it links fontconfig's own libraries, which a static fontconfig
keeps private.

The surface draws into an image and posts it to the window an activity is
given, asking the window which region has to be redrawn. The window bound to
that surface is given the screen, since a smaller one would sit in a corner.
Touches and keys become NSEvents. The surface is given up and taken back with
the activity.

The screen size is read from the bound window, or from GSAndroidScreenSize
where there is none.

Tests/cairo/androidpost and androidrepaint drive a real ANativeWindow through
AImageReader, so they need no activity.

### DIFF
--- a/Headers/android/AndroidServer.h
+++ b/Headers/android/AndroidServer.h
@@ -1,0 +1,133 @@
+/*
+   AndroidServer.h
+
+   Copyright (C) 2026 Free Software Foundation, Inc.
+
+   Author: Todd White <todd.white@thalion.global>
+   Date: August 2026
+
+   This file is part of the GNU Objective C Backend Library.
+
+   This library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Lesser General Public
+   License as published by the Free Software Foundation; either
+   version 2 of the License, or (at your option) any later version.
+
+   This library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Lesser General Public License for more details.
+
+   You should have received a copy of the GNU Lesser General Public
+   License along with this library; see the file COPYING.LIB.
+   If not, see <http://www.gnu.org/licenses/> or write to the
+   Free Software Foundation, 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110-1301, USA.
+*/
+
+#ifndef AndroidServer_h
+#define AndroidServer_h
+
+#include "config.h"
+
+#include <Foundation/NSGeometry.h>
+#include <Foundation/NSMapTable.h>
+#include <GNUstepGUI/GSDisplayServer.h>
+
+#include <android/input.h>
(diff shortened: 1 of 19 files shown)


Headers/android/AndroidServer.h
```diff
@@ -0,0 +1,73 @@
+/*
+   AndroidServer.h
+
+   Copyright (C) 2026 Free Software Foundation, Inc.
+
```
fredkiefer, line 5: You should put your name here

fredkiefer:
As I am no Android expert I would like to see @triplef comments on this PR.

triplef:
While we use GNUstep on Android, we only use libs-base (and libs-corebase) but leave the UI part to Qt, so I have no experience with libs-back or low-level graphics on Android and don’t feel in a position to comment on this implementation.

That being said it’s great to see this. Once it’s integrated we should add libs-back to [tools-android](https://github.com/gnustep/tools-android) as well (behind a `--with-gui` flag like on [tools-windows-msvc](https://github.com/gnustep/tools-windows-msvc)).

DTW-Thalion:
The reason I went down this rabbit hole was to try to address the Issue in libs-base asking for a CI Runner for Android. That is done, though it adds a lot of work to the CI which is another issue which @rfm rightly has pointed out needs solving.

That said, I presumed the point of doing so was to allow an iOS app to be ported to Android with API parity, and that of course would require a libs-gui / libs-back component, and so I went...

libs-back itself has received some changes now that makes each backend pretty much at functional parity with Cairo, and so having this done, it was fairly easy to cut and paste a template for Android, and subclass and implement the GState code, AndroidServer, etc.

What I can say is that if this really isn't something the project would want to **actively** support going forward, it is best to park it. The challenge is that the libs-back and libs-gui parts compile and run fine on Android now (the test suites) based on my changes, but there are a number of dependencies and environmental setups that are tricky for actual deployment.

I had to build an actual app to really test this and the app is running easily, but the setup to deploy an APK without using Android Studio, Java,etc.  and their tools is a fair bit of work, and while I think it can be automated with scripts, it is an effort not worth anybody's time unless it would be actually used.

If Qt is adequate, and there are no other users looking for direct iOS to Android porting solution, then we can let it be. If it is something we collectively think is worth it I will submit the harnesses for app building. 

We still will need to address the CI issue, as in whether we even test it in CI, maybe leaving it to a make check on install.

Headers/android/AndroidServer.h
DTW-Thalion: Added, and to the other files this adds. Source/android/GNUmakefile and its preamble come from the x11 backend's, so they keep Adam Fedor as well.

PR #235, opened 2026-08-09
CI: build and run the test suite on Android

DTW-Thalion:
libs-back is built and tested only on the systems it already runs on. This workflow cross-compiles the dependencies, tools-make, libs-base, libs-gui and libs-back for x86_64-linux-android at API 31 and runs Tests on an emulator.

The dependencies are cached on the hash of deps.sh, so the 9 minutes they take are paid only when it changes. tests.sh compiles each test, stages every library under the soname the loader asks for, deploys a GNUstep tree the bundle can be found in, and reads the result against expected-failures.txt.

The emulator image is cached on the ABI and the API level: creating it costs 86 to 116 seconds and restoring it 15. It is saved by a step of its own, since actions/cache writes nothing when the job has failed. A concurrency block cancels a run a newer push has superseded.

pkg-config is pointed at the cross prefix. Unset, it answers nothing on a runner with no host cairo, and every test including cairo.h fails to build while the rest reports healthy counts.

Measured against a branch carrying the android server: 286 assertions passed, 0 failed, nothing unbuilt, in about 4 minutes.

### DIFF
--- a/.github/scripts/android/android.cross.config
+++ b/.github/scripts/android/android.cross.config
@@ -1,0 +1,25 @@
+# libs-base cross answers for Android x86_64-linux-android, API 31.
+# Produced by running each configure test on an emulator.
+# NDK 29.0.14206865, libobjc2 -fobjc-runtime=gnustep-2.2.
+cross_CMDLINE_TERMINATED=1
+cross_NEED_WORD_ALIGNMENT=0
+cross_VASPRINTF_RETURNS_LENGTH=1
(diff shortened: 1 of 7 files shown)


fredkiefer:
I am really not sure whether we want even more test environments. You already had this discussion with Richard on libs-base and I would prefer to wait out the results from this. For me it would be better to have just three jobs running after each commit and have a separate weekly task running that tests more environments.

